PR #3, opened 2019-11-01
Add Moving Average Statistics Calculators

dabonnie:
This patch adds the utilities, and relevant tests, needed to make moving average statistics calculations.


system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,198 @@
...
+  std::shared_ptr<MovingAverageStatistics<double>> moving_average_statistics = nullptr;
+  int expected_count = 0;
+};
+
+// Useful testing constants
+const double TEST_DATA[] = {-3.5, -2.1, -1.1, 0.0, 4.7, 5.0, 6.7, 9.9, 11.0};
+const double EXPECTED_AVG = 3.4;
+const double EXPECTED_MIN = -3.5;
+const double EXPECTED_MAX = 11.0;
+const double EXPECTED_STD = 4.997999599839919955173;
```
piraka9011: What are the chances we get flakey tests on different OS b/c of precision?

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,198 @@
...
+  t2.join();
+  t3.join();
+
+  double control = static_cast<double>(total_sum.load()) / static_cast<double>(count.load());
+  double var = 0.00000000001;
+
+  ASSERT_NEAR(moving_average_statistics->average(), control, var);
+}
+
+TEST_F(MovingAverageStatisticsTestFixture, asdf) {
```
piraka9011: asdf? :)

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
dabonnie: > What are the chances we get flakey tests on different OS b/c of precision?

With the current tests and thresholds I believe low. I verified this by running the tests with 

```bash
'time ./test_moving_average_statistics --gtest_repeat=2000 --gtest_shuffle --gtest_break_on_failure'
```

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
dabonnie: Oops, nice catch and will fix.

dabonnie:
> Is this something we plan on extending with other stats or just a single moving average calculator?

Right now we have no immediate plan, but this originated from an internal code base that did just that.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+TEST_F(MovingAverageStatisticsTestFixture, test_minimum) {
+  for (double d : TEST_DATA) {
+    moving_average_statistics->add_item(d);
+    ASSERT_EQ(++expected_count, moving_average_statistics->get_count());
+  }
+
+  const double expected_minimum = -3.5;
+  EXPECT_EQ(moving_average_statistics->min(), expected_minimum);
```
prajakta-gokhale: Can use `EXPECTED_MINIMUM` here instead of local variable.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+TEST_F(MovingAverageStatisticsTestFixture, test_minimum_empty) {
+  ASSERT_TRUE(std::isnan(moving_average_statistics->min()));
+}
+
+TEST_F(MovingAverageStatisticsTestFixture, test_standard_deviation) {
+  for (double d : TEST_DATA) {
+    moving_average_statistics->add_item(d);
+  }
```
prajakta-gokhale: This code is duplicated multiple times. Maybe a utility function would be better?

system_metrics_collector/include/moving_average_statistics/types.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#ifndef MOVING_AVERAGE_STATISTICS__TYPES_HPP_
+#define MOVING_AVERAGE_STATISTICS__TYPES_HPP_
+
+/**
+ *  A container for statistics results
+ */
+struct StatisticResults
+{
+  double average;
```
prajakta-gokhale: I think it's better to initialize all the variables here.

dabonnie: Good idea, will do.

system_metrics_collector/CMakeLists.txt
```diff
@@ -28,15 +28,24 @@ find_package(rclcpp REQUIRED)
...
   ament_lint_auto_find_test_dependencies()
+
+  ament_add_gtest(test_moving_average_statistics
+          test/moving_average_statistics/test_moving_average_statistics.cpp)
+
+  install(TARGETS test_moving_average_statistics DESTINATION lib/${PROJECT_NAME})
+
+  if(TARGET test_multifile_reader)
+    target_link_libraries(test_multifile_reader system_metrics_collector)
+  endif()
```
thomas-moulard: You don't need this IF statement I think

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
thomas-moulard: This header is in include/ do we expect this class to be used from the outside?

system_metrics_collector/CMakeLists.txt
```diff
@@ -28,15 +28,24 @@ find_package(rclcpp REQUIRED)
 find_package(rcutils REQUIRED)
 find_package(metrics_statistics_msgs REQUIRED)
 
+include_directories(include)
```
thomas-moulard: Is this package intended to be a C++ library? If it isn't, you should put your headers in src/

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+ *
+ *  The statistics calculated are average, maximum, minimum, and standard deviation (population).
+ *  All are calculated online without storing the observation data. Specifically, the average is a running sum
+ *  and the variance is obtained by Welford's online algorithm
+ *  (reference: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford%27s_online_algorithm)
+ *  for standard deviation.
+ *
+ *  When statistics are not available, e.g. no observations have been made, NaNs are returned.
+ *
+ *  @tparam T the type to be observed. Note: this must conform to std::is_arithmetic.
```
thomas-moulard: Why can't we use always double here? That would simplify the code a lot and get rid of the template.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  }
+  /**
+   *  Return a StatisticResults object, containing average, minimum, maximum, standard deviation (population),
+   *  and sample count.
+   *  For the case of no observations, the average, min, max, and standard deviation are NaN.
+   *
+   *  @return StatisticResults object, containing average, minimum, maximum, standard deviation (population),
+   *  and sample count.
+  **/
+  StatisticResults getStatistics()
```
zmichaels11: Can this be ```const```?

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+template<
+  class T,
+  typename = typename std::enable_if<std::is_arithmetic<T>::value, T>::type>
+class MovingAverageStatistics
+{
+public:
+  /**
+   *  MovingAverageStatistics object constructor.
+  **/
```
thomas-moulard: Consider removing comments which aren't helpful.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  MovingAverageStatistics() = default;
+  ~MovingAverageStatistics() = default;
+
+  /**
+   *  Returns the arithmetic mean of all data recorded. If no observations have been made, returns NaN.
+   *
+   *  @return The arithmetic mean of all data recorded, or NaN if the sample count is 0.
+  **/
+  double average()
```
thomas-moulard:
```suggestion
  double average() const
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    return getStatistics().average;
+  }
+  /**
+   *  Returns the maximum value recorded. If size of list is zero, returns NaN.
+   *
+   *  @return The maximum value recorded, or NaN if size of data is zero.
+  **/
+  double max()
```
thomas-moulard:
```suggestion
  double max() const
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    return getStatistics().max;
+  }
+  /**
+   *  Returns the minimum value recorded. If size of list is zero, returns NaN.
+   *
+   *  @return The minimum value recorded, or NaN if size of data is zero.
+  **/
+  double min()
```
thomas-moulard:
```suggestion
  double min() const
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  }
+  /**
+   *  Returns the standard deviation (population) of all data recorded. If size of list is zero, returns NaN.
+   *
+   *  Variance is obtained by Welford's online algorithm,
+   *  see https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford%27s_online_algorithm
+   *
+   *  @return The standard deviation (population) of all data recorded, or NaN if size of data is zero.
+  **/
+  double standardDeviation()
```
thomas-moulard:
```suggestion
  double standardDeviation() const
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+   *  Returns the minimum value recorded. If size of list is zero, returns NaN.
+   *
+   *  @return The minimum value recorded, or NaN if size of data is zero.
+  **/
+  double min()
+  {
+    return getStatistics().min;
+  }
+  /**
+   *  Returns the standard deviation (population) of all data recorded. If size of list is zero, returns NaN.
```
thomas-moulard: 80 cols?

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    std::lock_guard<std::recursive_mutex> guard(mutex);
+    StatisticResults to_return;
+    to_return.sample_count = count_;
+
+    if (count_ == 0) {
+      to_return.average = nan("");
+      to_return.min = nan("");
+      to_return.max = nan("");
+      to_return.standard_deviation = nan("");
```
thomas-moulard: consider the more ""modern"" `std::nan` which is less sensitive to naming collisions

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  }
+  /**
+   *  Reset all calculated values. Equivalent to a new window for a moving average.
+  **/
+  void reset()
+  {
+    std::lock_guard<std::recursive_mutex> guard(mutex);
+    average_ = 0;
+    min_ = DBL_MAX;
+    max_ = DBL_MIN;
```
zmichaels11:
```c++
min_ = std::numeric_limits<double>::max();
max_ = std::numeric_limits<double>::min();  // might want lowest()
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+   *  Return a StatisticResults object, containing average, minimum, maximum, standard deviation (population),
+   *  and sample count.
+   *  For the case of no observations, the average, min, max, and standard deviation are NaN.
+   *
+   *  @return StatisticResults object, containing average, minimum, maximum, standard deviation (population),
+   *  and sample count.
+  **/
+  StatisticResults getStatistics()
+  {
+    std::lock_guard<std::recursive_mutex> guard(mutex);
```
thomas-moulard: You don't need a recursive_mutex here.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
dabonnie: Will do.

system_metrics_collector/CMakeLists.txt
dabonnie: You are correct: found it when adding another test in a separate branch. Will prune here.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
thomas-moulard: Consider using https://en.cppreference.com/w/cpp/types/numeric_limits instead as it is more portable than those older platform-specific C macros.

system_metrics_collector/include/moving_average_statistics/types.hpp
```diff
@@ -0,0 +1,30 @@
...
+  double min;
+  double max;
+  double standard_deviation;
+  int sample_count;
```
zmichaels11: You can probably increase `sample_count` to `long` without increasing `sizeof(StatisticResults)` due to padding.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+    average_ = 0;
+    min_ = DBL_MAX;
+    max_ = DBL_MIN;
+    sum_of_square_diff_from_mean_ = 0;
+    count_ = 0;
+  }
+  /**
+   *  Observe a sample for the given window. The input item is used to calculate statistics.
+   *
+   *  Derived class should override add_item() and call add_item_() with custom input.
```
thomas-moulard: Consider instead doing the following:

```
void add_item(const T & item) {
  on_item(item);
  add_item_(item);
}

virtual on_item(const T &) {}
```

This way the API can't be mis-used.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    add_item_(item);
+  }
+
+  /**
+   * Return the number of samples observed
+   *
+   * @return the number of samples observed
+   */
+  virtual int64_t get_count()
```
thomas-moulard:
```suggestion
  virtual int64_t get_count() const
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+  {
+    std::lock_guard<std::recursive_mutex> guard(mutex);
+    return count_;
+  }
+
+protected:
```
thomas-moulard: Consider making this private?

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: Hm, none of the linters have complained. I'll try rerunning and auto formatting.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+   *  Variance is obtained by Welford's online algorithm,
+   *  see https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford%27s_online_algorithm
+   *
+   *  @param item item to be stored.
+   */
+  void add_item_(const T & item)
+  {
+    std::lock_guard<std::recursive_mutex> guard(mutex);
+    count_++;
+    double previous_average_ = average_;
```
thomas-moulard:
```suggestion
    const double previous_average_ = average_;
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+    average_ = previous_average_ + (item - previous_average_) / count_;
+    min_ = std::min(min_, static_cast<double>(item));
+    max_ = std::max(max_, static_cast<double>(item));
+    sum_of_square_diff_from_mean_ = sum_of_square_diff_from_mean_ + (item - previous_average_) *
+      (item - average_);
+  }
+
+  mutable std::recursive_mutex mutex;
```
thomas-moulard:
```suggestion
  mutable std::mutex mutex;
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+private:
+  // cached values below
+  double average_ = 0;
+  double min_ = DBL_MAX;
+  double max_ = DBL_MIN;
+  double sum_of_square_diff_from_mean_ = 0;
+  int64_t count_ = 0;
```
thomas-moulard: I'd extract this into a separate struct for clarity.

```c++
struct StatisticsData {  // a better name would be preferable.
}
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: Nope, will fix.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
```diff
@@ -0,0 +1,189 @@
...
+   *  @param item item to be stored.
+   */
+  void add_item_(const T & item)
+  {
+    std::lock_guard<std::recursive_mutex> guard(mutex);
+    count_++;
+    double previous_average_ = average_;
+    average_ = previous_average_ + (item - previous_average_) / count_;
+    min_ = std::min(min_, static_cast<double>(item));
+    max_ = std::max(max_, static_cast<double>(item));
```
thomas-moulard: Let's drop the template if you're anyway going to cast everything back to double here, this presents little advantage.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
thomas-moulard: Consider calling this `copyStatistics` as this describes more accurately what you're doing here.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+  std::shared_ptr<MovingAverageStatistics<double>> moving_average_statistics = nullptr;
+  int expected_count = 0;
+};
+
+// Useful testing constants
+const double TEST_DATA[] = {-3.5, -2.1, -1.1, 0.0, 4.7, 5.0, 6.7, 9.9, 11.0};
+const double EXPECTED_AVG = 3.4;
+const double EXPECTED_MIN = -3.5;
+const double EXPECTED_MAX = 11.0;
+const double EXPECTED_STD = 4.997999599839919955173;
```
zmichaels11: Can these constants be `constexpr`?

Would it be possible to calculate `EXPECTED_AVG`, `EXPECTED_MIN`, `EXPECTED_MAX`, `EXPECTED_STD` from `TEST_DATA` using a constexpr expression?

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
thomas-moulard: edit: confused now, it seems like you already have a `StatisticResults` type - maybe just using it here would be better?

system_metrics_collector/include/moving_average_statistics/types.hpp
thomas-moulard: Consider keeping a 1:1 mapping between headers and types to facilitate the navigations. You could just call this `statistics_results.hpp`.

Regarding @prajakta-gokhale 's comment, the different is that keeping this w/o an inline initialization guarantee this type to be a POD struct, vs adding an inline initialization would make it non-POD.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <thread>
+#include <atomic>
```
thomas-moulard: alphasort

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+    moving_average_statistics = std::make_shared<MovingAverageStatistics<double>>();
+  }
+
+  void TearDown() override
+  {
+    moving_average_statistics->reset();
+  }
+
+protected:
+  std::shared_ptr<MovingAverageStatistics<double>> moving_average_statistics = nullptr;
```
thomas-moulard:
```suggestion
  std::unique_ptr<MovingAverageStatistics<double>> moving_average_statistics{};
```

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
zmichaels11: `std::array` has constexpr functions and may help in my second comment.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * Test fixture
+ */
+class MovingAverageStatisticsTestFixture : public ::testing::Test
+{
+public:
+  void SetUp() override
+  {
+    moving_average_statistics = std::make_shared<MovingAverageStatistics<double>>();
```
thomas-moulard:
```suggestion
    moving_average_statistics = std::make_unique<MovingAverageStatistics<double>>();
```

thomas-moulard: or maybe you'll need to do `moving_average_statistics.reset(new MovingAverageStatistics<double>());`

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+// Useful testing constants
+const double TEST_DATA[] = {-3.5, -2.1, -1.1, 0.0, 4.7, 5.0, 6.7, 9.9, 11.0};
+const double EXPECTED_AVG = 3.4;
+const double EXPECTED_MIN = -3.5;
+const double EXPECTED_MAX = 11.0;
+const double EXPECTED_STD = 4.997999599839919955173;
+const double EXPECTED_SIZE = 9;
+
+TEST_F(MovingAverageStatisticsTestFixture, sanity) {
+  ASSERT_TRUE(true);
```
thomas-moulard: remove

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+  ASSERT_NE(moving_average_statistics, nullptr);
+}
+
```
thomas-moulard: remove? this does not sound super helpful

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+TEST_F(MovingAverageStatisticsTestFixture, test_average) {
+  for (double d : TEST_DATA) {
+    moving_average_statistics->add_item(d);
+    ASSERT_EQ(++expected_count, moving_average_statistics->get_count());
```
thomas-moulard:
```suggestion
    EXPECT_EQ(++expected_count, moving_average_statistics->get_count());
```

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+  int expected_count = 0;
+};
+
+// Useful testing constants
+const double TEST_DATA[] = {-3.5, -2.1, -1.1, 0.0, 4.7, 5.0, 6.7, 9.9, 11.0};
+const double EXPECTED_AVG = 3.4;
+const double EXPECTED_MIN = -3.5;
+const double EXPECTED_MAX = 11.0;
+const double EXPECTED_STD = 4.997999599839919955173;
+const double EXPECTED_SIZE = 9;
```
thomas-moulard: nit: top-level global constants should be declared as static.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+    });
+  std::thread t2([this, &count, &total_sum]() {
+      for (double i = 1; i < 2101.0; i++) {
+        moving_average_statistics->add_item(i);
+        count++;
+        total_sum += i;
+      }
+    });
+  std::thread t3([this, &count, &total_sum]() {
+      for (double i = 1; i < 3101.0; i++) {
```
thomas-moulard: it is preferable to index loops by int and cast to double when calling `add_item`

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
```diff
@@ -0,0 +1,194 @@
...
+        total_sum += i;
+      }
+    });
+
+  t1.join();
+  t2.join();
+  t3.join();
+
+  double control = static_cast<double>(total_sum.load()) / static_cast<double>(count.load());
+  double var = 0.00000000001;
```
thomas-moulard:
```suggestion
  double var = 1e-16;
```

You are very close to machine precision - are you sure it's going to work?

zmichaels11: If it is close to machine epsilon, maybe just use 
```c++
std::numeric_limits<double>::epsilon()
```

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: It's not quite a copy: note how the std_dev is calculated at this point.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: I agree. We had the case before to try and extend (with multiple stats collectors), but I don't see how that use case applies here anymore.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
zmichaels11: FYI: You will have to define `NOMINMAX` in `CMakeLists.txt` for Windows. Otherwise the Windows compiler may use the macros

dabonnie: Right now we aren't supporting Windows: main use case is for linux.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
dabonnie: Assert was used intentionally, as there's no reason to continue the test for the early failure.

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
dabonnie: > remove? this does not sound super helpful

I disagree. For every fixture I always add a sanity test for the fixture itself. If any changes are made to the fixture or its member classes and this fails (using gtest flags to run individual tests) then it helps pinpoint where issues may live. 

I've found this extremely useful in past projects.

system_metrics_collector/include/moving_average_statistics/types.hpp
dabonnie: This actually needs to be int64_t in order to match the observed value.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: Yes

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: Since we're removing templating and don't have a use case for inheritance I don't see the value in the current approach.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: At some point we will want to do so, but moving to src for now while we develop.

system_metrics_collector/CMakeLists.txt
dabonnie: At some point we will want to do so, but moving to src for now while we develop.

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: Since we're caching the sum of square difference from the mean I found it clearer to only return the statistics struct when it's requested

system_metrics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp
dabonnie: I'm having some issues with this, maybe we can sit down or punt to another PR.

zmichaels11: ok

dabonnie:
> Thanks for removing the template, do you mind splitting the file in a CPP and a HPP ?

@thomas-moulard  how do you envision the structure if we want the header to be private?

system_metrics_collector/include/moving_average_statistics/moving_average.hpp
dabonnie: Moved to src

PR #5, opened 2019-11-05
Add github actions lint and test

dabonnie:


dabonnie:
@thomas-moulard it's unclear to me why the build is failing with the current CMakeLists.txt.

zmichaels11:
@dabonnie 
```
Cannot find source file:

```

thomas-moulard:
I merged master, that should fix this issue.

thomas-moulard:
@dabonnie Merging that in as it's better than nothing, seems like getting coverage for C++ needs some more work.

PR #6, opened 2019-11-07
Add periodic API

dabonnie:
Add relevant class and abstractions to perform periodic measurements.

Note: the Collector and Periodic patter follows previous work from ROS CloudWatch CEs and ROS2 QoS System tests.

### DIFF
--- a/system_metrics_collector/src/system_metrics_collector/collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.cpp
@@ -23,7 +23,7 @@
 
 bool Collector::start()
 {
-  std::unique_lock<std::recursive_mutex> ulock(mutex);
+  std::unique_lock<std::mutex> ulock(mutex);
   if (started_) {
     return false;
   }
@@ -33,7 +33,7 @@ bool Collector::start()
 
 bool Collector::stop()
 {
-  std::unique_lock<std::recursive_mutex> ulock(mutex);
+  std::unique_lock<std::mutex> ulock(mutex);
   if (!started_) {
     return false;
   }
@@ -61,7 +61,7 @@ void Collector::clearCurrentMeasurements()
 
 bool Collector::isStarted() const
 {
-  std::unique_lock<std::recursive_mutex> ulock(mutex);
+  std::unique_lock<std::mutex> ulock(mutex);
   return started_;
 }
 
--- a/system_metrics_collector/src/system_metrics_collector/collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.cpp
@@ -12,14 +12,39 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "collector.hpp"
+
+#include <mutex>
+#include <sstream>
+#include <string>
 
 #include "../../src/moving_average_statistics/moving_average.hpp"
 #include "../../src/moving_average_statistics/types.hpp"
 
-#include "collector.hpp"
+bool Collector::start()
+{
+  std::unique_lock<std::recursive_mutex> ulock(mutex);
+  if (started_) {
+    return false;
+  }
+  started_ = true;
+  return setupStart();
+}
 
+bool Collector::stop()
+{
+  std::unique_lock<std::recursive_mutex> ulock(mutex);
+  if (!started_) {
+    return false;
+  }
+  started_ = false;
 
-void Collector::acceptData(double measurement)
+  const bool ret = setupStop();
+  clearCurrentMeasurements();
+  return ret;
+}
+
+void Collector::acceptData(const double measurement)
 {
   collected_data_.addMeasurement(measurement);
 }
@@ -32,4 +57,18 @@ StatisticData Collector::getStatisticsResults() const
 void Collector::clearCurrentMeasurements()
 {
   collected_data_.reset();
+}
+
+bool Collector::isStarted() const
+{
+  std::unique_lock<std::recursive_mutex> ulock(mutex);
+  return started_;
+}
+
+std::string Collector::getStatusString()
+{
+  std::stringstream ss;
+  ss << "started=" << (isStarted() ? "true" : "false") <<
+    ", " << statisticsDataToString(getStatisticsResults());
+  return ss.str();
 }
--- a/system_metrics_collector/src/system_metrics_collector/collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.cpp
@@ -33,13 +33,16 @@ bool Collector::start()
 
 bool Collector::stop()
 {
-  std::unique_lock<std::mutex> ulock(mutex);
-  if (!started_) {
-    return false;
-  }
-  started_ = false;
+  bool ret = false;
+  {
+    std::unique_lock<std::mutex> ulock(mutex);
+    if (!started_) {
+      return false;
+    }
+    started_ = false;
 
-  const bool ret = setupStop();
+    ret = setupStop();
+  }
   clearCurrentMeasurements();
   return ret;
 }
@@ -65,7 +68,7 @@ bool Collector::isStarted() const
   return started_;
 }
 
-std::string Collector::getStatusString()
+std::string Collector::getStatusString() const
 {
   std::stringstream ss;
   ss << "started=" << (isStarted() ? "true" : "false") <<
--- a/system_metrics_collector/src/system_metrics_collector/collector.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.hpp
@@ -15,11 +15,14 @@
 #ifndef SYSTEM_METRICS_COLLECTOR__COLLECTOR_HPP_
 #define SYSTEM_METRICS_COLLECTOR__COLLECTOR_HPP_
 
+#include <mutex>
+#include <string>
+
 #include "../moving_average_statistics/moving_average.hpp"
 #include "../moving_average_statistics/types.hpp"
 
 /**
- * Simple wrapping class in order to collect observed data and generate statistics for the given observations.
+ * Simple class in order to collect observed data and generate statistics for the given observations.
  */
 class Collector
 {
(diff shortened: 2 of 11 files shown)


system_metrics_collector/src/system_metrics_collector/collector.hpp
```diff
@@ -28,26 +31,45 @@ class Collector
...
+
+  /**
+   * Override in order to perform necessary starting steps. Overridden methods should lock the recursive mutex class
+   * member 'mutex'
    *
-   * @return true if started, false if an error occured
+   * @return true if setup was successful, false otherwise.
    */
-  virtual bool start() = 0;
+  virtual bool setupStart() = 0;
```
thomas-moulard: - Could we move this protected?
- Let's rename this to `setupStartLocked` and document the mutex need to be locked before the function is being called

system_metrics_collector/src/system_metrics_collector/collector.hpp
```diff
@@ -28,26 +31,45 @@ class Collector
...
-  virtual bool stop() = 0;
+  bool stop();
+
+  /**
+   * Override in order to perform necessary teardown. Overridden methods should lock the recursive mutex class
+   * member 'mutex'
+   *
+   * @return true if teardown was successful, false otherwise.
+   */
+  virtual bool setupStop() = 0;
```
thomas-moulard: ditto

system_metrics_collector/src/system_metrics_collector/collector.hpp
```diff
@@ -61,6 +83,25 @@ class Collector
...
+   * Return a pretty printed status representation of this class
+   *
+   * @return a string detailing the current status
+   */
+  virtual std::string getStatusString();
+
+  // todo @dabonnie uptime (once start has been called)
+
+protected:
+  mutable std::recursive_mutex mutex;  // recursive for child classes
```
thomas-moulard:
```suggestion
  mutable std::mutex mutex;
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -0,0 +1,75 @@
...
+  const std::chrono::milliseconds & measurement_period,
+  const std::string & publishing_topic)
+: Node(name),
+  measurement_period_(measurement_period),
+  publishing_topic_(publishing_topic)
+{}
+
+bool PeriodicMeasurementNode::setupStart()
+{
+  std::unique_lock<std::recursive_mutex> ulock(mutex);
```
thomas-moulard:
```suggestion
  std::unique_lock<std::mutex> ulock(mutex);
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -0,0 +1,75 @@
...
+    measurement_timer_->cancel();
+    measurement_timer_ = nullptr;
+  }
+  return true;
+}
+
+void PeriodicMeasurementNode::performPeriodicMeasurement()
+{
+  std::unique_lock<std::recursive_mutex> ulock(mutex);
+  periodicMeasurement();
```
thomas-moulard:
```suggestion
  periodicMeasurementLocked();
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -0,0 +1,87 @@
...
+{
+public:
+  /**
+   * Construct a PeriodicMeasurementNode.
+   *
+   * @param name the name of this node
+   * @param topic the topic for publishing data
+   * @param measurement_period
+   */
+  explicit PeriodicMeasurementNode(
```
thomas-moulard:
```suggestion
  PeriodicMeasurementNode(
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Override this method to perform a measurement. This is called via performPeriodicMeasurement
+   * with the period defined in the constructor.
+   */
+  virtual void periodicMeasurement() = 0;
+
+private:
+  std::string publishing_topic_;
+  std::chrono::milliseconds measurement_period_;
+  rclcpp::TimerBase::SharedPtr measurement_timer_{nullptr};
```
thomas-moulard:
```suggestion
  rclcpp::TimerBase::SharedPtr measurement_timer_;
```

Either everything is initialized inline (and ctor = default), or nothing (and we need a ctor). Doing half is not consistent.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -0,0 +1,75 @@
...
+  return true;
+}
+
+bool PeriodicMeasurementNode::setupStop()
+{
+  std::unique_lock<std::recursive_mutex> ulock(mutex);
+
+  if (measurement_timer_) {
+    measurement_timer_->cancel();
+    measurement_timer_ = nullptr;
```
thomas-moulard: (nit)

```suggestion
    measurement_timer_.reset();
```

system_metrics_collector/src/system_metrics_collector/collector.hpp
dabonnie: Thanks for the articles above! 

I think the setup methods should be moved to private. Doing so allows us to use a single mutex in the collector class and removing the locking behavior in the periodic_measurement_node class.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: I don't think we need this lock anymore: especially since the collection addition (via the moving average container) is lock protected.

dabonnie:
Pushed changes addressing design, but will add thread safety annotations in this PR.

system_metrics_collector/src/system_metrics_collector/collector.cpp
```diff
@@ -33,3 +58,17 @@ void Collector::clearCurrentMeasurements()
...
   collected_data_.reset();
 }
+
+bool Collector::isStarted() const
+{
+  std::unique_lock<std::mutex> ulock(mutex);
+  return started_;
+}
+
+std::string Collector::getStatusString()
```
zmichaels11:
```c++
std::string Collector::getStatusString() const
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -0,0 +1,66 @@
...
+bool PeriodicMeasurementNode::setupStop()
+{
+  if (measurement_timer_) {
+    measurement_timer_->cancel();
+    measurement_timer_.reset();
+  }
+  return true;
+}
+
+std::string PeriodicMeasurementNode::getStatusString()
```
zmichaels11:
```c++
std::string PeriodicMeasurementNode::getStatusString() const
```

system_metrics_collector/src/system_metrics_collector/collector.cpp
```diff
@@ -12,14 +12,39 @@
...
+bool Collector::stop()
+{
+  std::unique_lock<std::mutex> ulock(mutex);
+  if (!started_) {
+    return false;
+  }
+  started_ = false;
 
+  const bool ret = setupStop();
+  clearCurrentMeasurements();
```
thomas-moulard: does that need to be a single lock?

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -0,0 +1,142 @@
...
+  // made public to manually test
+  void periodicMeasurement() override
+  {
+    sum += 1;
+    acceptData(static_cast<double>(sum.load()));
+  }
+
+private:
+  std::atomic<int> sum{0};
+  std::condition_variable_any cv;
```
thomas-moulard:
```suggestion
  std::condition_variable_any cv{};
```

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "../../src/moving_average_statistics/types.hpp"
+
+/**
+ * Simple extension to test basic functionality
+ */
+class TestPeriodicMeasurementNode : public PeriodicMeasurementNode
+{
+public:
+  TestPeriodicMeasurementNode(
+    const std::string name,
```
thomas-moulard:
```suggestion
    const std::string & name,
```

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
dabonnie: This is actually leftover testing cruft, removing.

system_metrics_collector/src/system_metrics_collector/collector.cpp
dabonnie: Do you mean that clearCurrentMeasurements() doesn't need to be locked in this scope, because it's already protected? That makes sense to me, while keeping setupStop in the locked scope as a private method.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Simple extension to test basic functionality
+ */
+class TestPeriodicMeasurementNode : public PeriodicMeasurementNode
+{
+public:
+  TestPeriodicMeasurementNode(
+    const std::string & name,
+    const std::chrono::milliseconds & measurement_period,
```
zmichaels11:
```c++
const std::chrono::miliseconds measurement_period
```
sizeof(std::chrono::milliseconds) is 8, so passing by ref might cost more than passing by val
http://cpp.sh/2dndi

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <string>
+#include <memory>
+#include <mutex>
+
+#include "../../src/system_metrics_collector/collector.hpp"
+#include "../../src/system_metrics_collector/periodic_measurement_node.hpp"
+#include "../../src/moving_average_statistics/types.hpp"
+
+
+static constexpr const std::chrono::milliseconds TEST_LENGTH =
```
zmichaels11:
```c++
namespace
{
  constexpr const auto TEST_LENGTH = std::chrono::milliseconds(250);
  constexpr const auto TEST_PERIOD = std::chrono::milliseconds(50);
}
```

[C++ core guidelines suggests using anonymous namespaces for non-exported entities](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnonexported-entities)

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
dabonnie: I'm going to change all the constructors to take a val instead of ref. This follows the ROS2 timer definitions as well.

PR #11, opened 2019-11-12
Measure linux cpu active percentage

dabonnie:
This PR adds a node which periodically measures the system active CPU %. This can be manually verified, on a linux system, with the example main function. Here's sample output with a window reset:

```
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=7.551427, min=1.282051, max=47.272727, std_dev=9.692852, count=54
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 9.859155
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=7.593386, min=1.282051, max=47.272727, std_dev=9.609279, count=55
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 2.631579
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=7.504782, min=1.282051, max=47.272727, std_dev=9.545739, count=56
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 16.666667
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=7.665517, min=1.282051, max=47.272727, std_dev=9.537783, count=57
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 1.298701
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=7.555745, min=1.282051, max=47.272727, std_dev=9.491455, count=58
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 11.111111
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=7.616005, min=1.282051, max=47.272727, std_dev=9.421859, count=59
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 13.043478
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=7.706463, min=1.282051, max=47.272727, std_dev=9.368814, count=60
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 15.942029
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=15.942029, min=15.942029, max=15.942029, std_dev=0.000000, count=1
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 5.194805
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=10.568417, min=5.194805, max=15.942029, std_dev=5.373612, count=2
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 3.846154
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=8.327663, min=3.846154, max=15.942029, std_dev=5.412248, count=3
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 2.564103
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=6.886773, min=2.564103, max=15.942029, std_dev=5.310161, count=4
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 3.846154
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=6.278649, min=2.564103, max=15.942029, std_dev=4.902806, count=5
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 17.910448
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=8.217282, min=2.564103, max=17.910448, std_dev=6.230791, count=6
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 1.298701
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=7.228913, min=1.298701, max=17.910448, std_dev=6.256027, count=7
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 2.564103
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=6.645812, min=1.298701, max=17.910448, std_dev=6.051917, count=8
[DEBUG] [linuxCpuCollector]: performPeriodicMeasurement: 2.564103
[DEBUG] [linuxCpuCollector]: name=linuxCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, clear_measurements_on_publish_=1, started=true, avg=6.192289, min=1.298701, max=17.910448, std_dev=5.848217, count=9
```

### DIFF
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -34,7 +34,7 @@ add_library(system_metrics_collector SHARED
   src/system_metrics_collector/collector.cpp
   src/system_metrics_collector/periodic_measurement_node.cpp
   src/system_metrics_collector/linux_cpu_measurement_node.cpp
-  src/system_metrics_collector/cpu_data.hpp
+  src/system_metrics_collector/cpu_data.cpp
 )
 
 ament_target_dependencies(system_metrics_collector rclcpp rcpputils)
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -33,7 +33,7 @@ add_library(system_metrics_collector SHARED
   src/moving_average_statistics/types.hpp
   src/system_metrics_collector/collector.cpp
   src/system_metrics_collector/periodic_measurement_node.cpp
-  src/system_metrics_collector/linux_cpu_measurement_node.hpp
+  src/system_metrics_collector/linux_cpu_measurement_node.cpp
   src/system_metrics_collector/cpu_data.hpp
 )
 
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -32,10 +32,17 @@ add_library(system_metrics_collector SHARED
   src/moving_average_statistics/moving_average.cpp
   src/moving_average_statistics/types.hpp
   src/system_metrics_collector/collector.cpp
-  src/system_metrics_collector/periodic_measurement_node.cpp)
+  src/system_metrics_collector/periodic_measurement_node.cpp
+  src/system_metrics_collector/linux_cpu_measurement_node.hpp
+  src/system_metrics_collector/cpu_data.hpp
+)
 
 ament_target_dependencies(system_metrics_collector rclcpp rcpputils)
 ament_export_libraries(system_metrics_collector)
+
+add_executable(main src/system_metrics_collector/main.cpp)
+target_link_libraries(main system_metrics_collector)
+ament_target_dependencies(main)
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
@@ -57,6 +64,11 @@ if(BUILD_TESTING)
     test/system_metrics_collector/test_periodic_measurement_node.cpp)
   target_link_libraries(test_periodic_measurement_node system_metrics_collector)
   ament_target_dependencies(test_periodic_measurement_node rclcpp)
+
+  ament_add_gtest(test_linux_cpu_measurement_node
+          test/system_metrics_collector/test_linux_cpu_measurement.cpp)
+  target_link_libraries(test_linux_cpu_measurement_node system_metrics_collector)
+  ament_target_dependencies(test_linux_cpu_measurement_node rclcpp)
 
   install(TARGETS
     test_moving_average_statistics
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -34,9 +34,9 @@ find_package(metrics_statistics_msgs REQUIRED)
 
 add_library(system_metrics_collector SHARED
   src/system_metrics_collector/collector.cpp
-  src/system_metrics_collector/cpu_data.cpp
   src/system_metrics_collector/linux_cpu_measurement_node.cpp
   src/system_metrics_collector/periodic_measurement_node.cpp
+  src/system_metrics_collector/proc_cpu_data.cpp
   src/moving_average_statistics/moving_average.cpp
   src/moving_average_statistics/types.hpp
 )
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -22,6 +22,10 @@ foreach(FLAG ${FLAGS})
   endif()
 endforeach()
 
+if(WIN32)
+  add_definitions(-DNOMINMAX)
+endif()
+
 # find dependencies
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
@@ -29,12 +33,12 @@ find_package(rcutils REQUIRED)
 find_package(metrics_statistics_msgs REQUIRED)
 
 add_library(system_metrics_collector SHARED
+  src/system_metrics_collector/collector.cpp
+  src/system_metrics_collector/cpu_data.cpp
+  src/system_metrics_collector/linux_cpu_measurement_node.cpp
+  src/system_metrics_collector/periodic_measurement_node.cpp
   src/moving_average_statistics/moving_average.cpp
   src/moving_average_statistics/types.hpp
-  src/system_metrics_collector/collector.cpp
-  src/system_metrics_collector/periodic_measurement_node.cpp
-  src/system_metrics_collector/linux_cpu_measurement_node.cpp
-  src/system_metrics_collector/cpu_data.cpp
 )
 
 ament_target_dependencies(system_metrics_collector rclcpp rcpputils)
@@ -50,25 +54,25 @@ if(BUILD_TESTING)
 
   ament_lint_auto_find_test_dependencies()
 
-  ament_add_gtest(test_moving_average_statistics
-    test/moving_average_statistics/test_moving_average_statistics.cpp)
-  target_link_libraries(test_moving_average_statistics system_metrics_collector)
-  ament_target_dependencies(test_moving_average_statistics rclcpp)
-
   ament_add_gtest(test_collector
     test/system_metrics_collector/test_collector.cpp)
   target_link_libraries(test_collector system_metrics_collector)
   ament_target_dependencies(test_collector rclcpp)
 
-  ament_add_gtest(test_periodic_measurement_node
-    test/system_metrics_collector/test_periodic_measurement_node.cpp)
-  target_link_libraries(test_periodic_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_periodic_measurement_node rclcpp)
-
   ament_add_gtest(test_linux_cpu_measurement_node
           test/system_metrics_collector/test_linux_cpu_measurement.cpp)
   target_link_libraries(test_linux_cpu_measurement_node system_metrics_collector)
   ament_target_dependencies(test_linux_cpu_measurement_node rclcpp)
+
+  ament_add_gtest(test_moving_average_statistics
+          test/moving_average_statistics/test_moving_average_statistics.cpp)
+  target_link_libraries(test_moving_average_statistics system_metrics_collector)
+  ament_target_dependencies(test_moving_average_statistics rclcpp)
+
+  ament_add_gtest(test_periodic_measurement_node
+    test/system_metrics_collector/test_periodic_measurement_node.cpp)
+  target_link_libraries(test_periodic_measurement_node system_metrics_collector)
+  ament_target_dependencies(test_periodic_measurement_node rclcpp)
 
   install(TARGETS
     test_moving_average_statistics
--- a/system_metrics_collector/src/moving_average_statistics/moving_average.cpp
+++ b/system_metrics_collector/src/moving_average_statistics/moving_average.cpp
@@ -77,13 +77,16 @@ void MovingAverageStatistics::reset()
 void MovingAverageStatistics::addMeasurement(const double item)
 {
   std::lock_guard<std::mutex> guard(mutex);
-  count_++;
-  const double previous_average_ = average_;
-  average_ = previous_average_ + (item - previous_average_) / count_;
-  min_ = std::min(min_, item);
-  max_ = std::max(max_, item);
-  sum_of_square_diff_from_mean_ = sum_of_square_diff_from_mean_ + (item - previous_average_) *
-    (item - average_);
+
+  if (!std::isnan(item)) {
+    count_++;
+    const double previous_average_ = average_;
+    average_ = previous_average_ + (item - previous_average_) / count_;
+    min_ = std::min(min_, item);
+    max_ = std::max(max_, item);
+    sum_of_square_diff_from_mean_ = sum_of_square_diff_from_mean_ + (item - previous_average_) *
+      (item - average_);
+  }
 }
 
 uint64_t MovingAverageStatistics::getCount() const
--- a/system_metrics_collector/src/moving_average_statistics/types.hpp
+++ b/system_metrics_collector/src/moving_average_statistics/types.hpp
@@ -37,7 +37,7 @@ struct StatisticData
  * @return std::string formatted struct contents in an easily readable format, e.g.,
  * /"avg=1, min=2, max=3, std_dev=4, count=5/"
  */
-static std::string statisticsDataToString(StatisticData results)
+static std::string statisticsDataToString(const StatisticData & results)
 {
   std::stringstream ss;
   ss << "avg=" << std::to_string(results.average) << ", min=" << std::to_string(results.min) <<
--- a/system_metrics_collector/src/system_metrics_collector/collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.cpp
@@ -17,7 +17,7 @@
 #include <mutex>
 #include <sstream>
 #include <string>
-#include <iostream>
+
 #include "../../src/moving_average_statistics/moving_average.hpp"
 #include "../../src/moving_average_statistics/types.hpp"
 
--- a/system_metrics_collector/src/system_metrics_collector/collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.cpp
@@ -17,7 +17,7 @@
 #include <mutex>
 #include <sstream>
 #include <string>
-
+#include <iostream>
 #include "../../src/moving_average_statistics/moving_average.hpp"
 #include "../../src/moving_average_statistics/types.hpp"
 
--- a/system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
@@ -15,36 +15,36 @@
 #ifndef SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
 #define SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
 
-#include <string>
+#include <array>
 #include <sstream>
+#include <string>
+
+/**
+ * Enum representing each item in a /proc/stat cpu line
+ */
+enum class ProcCpuStates
+{
+  kUser = 0,
+  kNice,
+  kSystem,
+  kIdle,
+  kIOWait,
+  kIrq,
+  kSoftIrq,
+  kSteal,
+  kNumProcCpuStates
+};
 
 /**
  * Struct containing data parsed from a /proc/stat cpu line
  */
-struct ProcCpuData
+class ProcCpuData
 {
-  static constexpr const char EMPTY_LABEL[] = "empty";
-  /**
-   * Enum representing each item in a /proc/stat cpu line
-   */
-  enum ProcCpuStates
-  {
-    user = 0,
-    nice,
-    system,
-    idle,
-    iOWait,
-    irq,
-    softIrq,
-    steal,
-    kNumProcCpuStates
-  };
+public:
+  ProcCpuData() = default;
+  virtual ~ProcCpuData() = default;
 
-  /**
-   * The cpu label of the line parsed
-   */
-  std::string cpu_label = EMPTY_LABEL;
-  size_t times[kNumProcCpuStates] = {0};
+  static constexpr const char EMPTY_LABEL[] = "empty";
 
   /**
    * Return the idle time
@@ -73,6 +73,17 @@ struct ProcCpuData
    * @return true if empty (no valid data), false otherwise.
    */
   bool isMeasurementEmpty() const;
+
+  /**
+   * The cpu label of the line parsed
+   */
+  std::string cpu_label{EMPTY_LABEL};
+
+  /**
+   * Array contained the parsed CPU data, where each index
+   * of ProcCpuStates contains its labeled data.
+   */
+  std::array<int, static_cast<int>(ProcCpuStates::kNumProcCpuStates)> times{};
 };
 
 #endif  // SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
--- a/system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
@@ -1,0 +1,86 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
+#define SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
+
+#include <string>
+#include <sstream>
+
+/**
+ * Struct containing data parsed from a /proc/stat cpu line
+ */
+struct ProcCpuData
+{
+  /**
+   * Enum representing each item in a /proc/stat cpu line
+   */
+  enum ProcCpuStates
+  {
+    user = 0,
+    nice,
+    system,
+    idle,
+    iOWait,
+    irq,
+    softIrq,
+    steal,
+    kNumProcCpuStates
+  };
+
+  /**
+   * The cpu label of the line parsed
+   */
+  std::string cpu_label = "empty";  // todo fixme magic string constant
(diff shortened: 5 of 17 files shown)


system_metrics_collector/src/moving_average_statistics/types.hpp
```diff
@@ -37,7 +37,7 @@ struct StatisticData
  * @return std::string formatted struct contents in an easily readable format, e.g.,
  * /"avg=1, min=2, max=3, std_dev=4, count=5/"
  */
-static std::string statisticsDataToString(StatisticData results)
+static std::string statisticsDataToString(const StatisticData & results)
```
zmichaels11: This copies the function into the local file scope, which then gets inlined.

I'm not 100% sure here, but I think its more proper to use the modifier `inline` since the intent is that the function is to be copied in.

If the intent is that the function should be non-exportable, it should instead be declared inside an anonymous namespace.

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Struct containing data parsed from a /proc/stat cpu line
+ */
+struct ProcCpuData
+{
+  static constexpr const char EMPTY_LABEL[] = "empty";
+  /**
+   * Enum representing each item in a /proc/stat cpu line
+   */
+  enum ProcCpuStates
```
zmichaels11:
```c++
enum class ProcCpuStates
```

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -0,0 +1,78 @@
...
+   */
+  size_t getActiveTime() const;
+
+  /**
+   * Return a pretty printed string of the ProcCpuData struct.
+   *
+   * @param data the struct to print
+   * @return a formatted string of the input struct
+   */
+  std::string toString();
```
zmichaels11:
```c++
std::string toString() const
```
?

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+  const double active_time = measurement2.getActiveTime() - measurement1.getActiveTime();
+  const double total_time = (measurement2.getIdleTime() + measurement2.getActiveTime()) -
+    (measurement1.getIdleTime() + measurement2.getActiveTime());
+
+  return 100.0 * active_time / total_time;
+}
+
+double LinuxCpuMeasurementNode::periodicMeasurement()
+{
+  auto measurement1 = makeSingleMeasurement();
```
zmichaels11:
```c++
const auto measurement1 = ...
const auto measurement2 = ...
const auto cpu_percentage = ...
```

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  return cpu_percentage;
+}
+
+ProcCpuData LinuxCpuMeasurementNode::makeSingleMeasurement()
+{
+  std::ifstream stat_file(PROC_STAT_FILE);
+  std::string line;
+  std::getline(stat_file, line);
+  stat_file.close();
```
zmichaels11: `ifstream::close` is handled by its deconstructor.
You might want to check `ifstream::good` before reading though.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+  virtual ~LinuxCpuMeasurementNode() = default;
+
+  /**
+   * Parse a line read from /proc/stat
+   *
+   * @param stat_cpu_line a line from /proc/stat
+   * @return ProcCpuData struct populated if parsed, otherwise empty
+   */
+  static ProcCpuData processLine(const std::string & stat_cpu_line);
```
zmichaels11: does it make sense to make this a free function inside `linux_cpu_measurement_node.cpp`?

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @param argc
+ * @param argv
+ * @return
+ */
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+
+  std::shared_ptr<LinuxCpuMeasurementNode> node = std::make_shared<LinuxCpuMeasurementNode>(
```
zmichaels11:
```c++
auto node = std::make_shared<LinuxCpuMeasurementNode>(...)
```

system_metrics_collector/test/system_metrics_collector/test_linux_cpu_measurement.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <iostream>
+#include <string>
+#include <memory>
+#include <mutex>
+
+#include "../../src/system_metrics_collector/linux_cpu_measurement_node.hpp"
+
+namespace
+{
+static constexpr const char proc_sample_1[] =
```
zmichaels11: nit: `static` is redundant here.

system_metrics_collector/src/moving_average_statistics/types.hpp
dabonnie: I'll leave it alone for now as discussed in person.

system_metrics_collector/CMakeLists.txt
```diff
@@ -32,11 +32,18 @@ add_library(system_metrics_collector SHARED
   src/moving_average_statistics/moving_average.cpp
   src/moving_average_statistics/types.hpp
   src/system_metrics_collector/collector.cpp
-  src/system_metrics_collector/periodic_measurement_node.cpp)
+  src/system_metrics_collector/periodic_measurement_node.cpp
+  src/system_metrics_collector/linux_cpu_measurement_node.cpp
+  src/system_metrics_collector/cpu_data.cpp
```
thomas-moulard: alphasort

system_metrics_collector/CMakeLists.txt
```diff
@@ -58,6 +65,11 @@ if(BUILD_TESTING)
   target_link_libraries(test_periodic_measurement_node system_metrics_collector)
   ament_target_dependencies(test_periodic_measurement_node rclcpp)
 
+  ament_add_gtest(test_linux_cpu_measurement_node
+          test/system_metrics_collector/test_linux_cpu_measurement.cpp)
+  target_link_libraries(test_linux_cpu_measurement_node system_metrics_collector)
+  ament_target_dependencies(test_linux_cpu_measurement_node rclcpp)
```
thomas-moulard: alphasort tests

system_metrics_collector/src/system_metrics_collector/collector.cpp
```diff
@@ -17,7 +17,7 @@
 #include <mutex>
 #include <sstream>
 #include <string>
-
+#include <iostream>
```
thomas-moulard: alphasort

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -0,0 +1,78 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
+#define SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
+
+#include <string>
+#include <sstream>
```
thomas-moulard: alphasort

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
thomas-moulard:
```suggestion
  enum class ProcCpuStates
```

Prefer enum class whenever possible (enum is older variant inherited from C)

https://www.geeksforgeeks.org/enum-classes-in-c-and-their-advantage-over-enum-datatype/

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -0,0 +1,78 @@
...
+  enum ProcCpuStates
+  {
+    user = 0,
+    nice,
+    system,
+    idle,
+    iOWait,
+    irq,
+    softIrq,
+    steal,
```
thomas-moulard: Please follow the ROS 2 coding style:
https://google.github.io/styleguide/cppguide.html#Enumerator_Names

user -> kUser, etc.

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -0,0 +1,78 @@
...
+    kNumProcCpuStates
+  };
+
+  /**
+   * The cpu label of the line parsed
+   */
+  std::string cpu_label = EMPTY_LABEL;
```
thomas-moulard:
```suggestion
  std::string cpu_label{EMPTY_LABEL};
```

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -0,0 +1,78 @@
...
+  size_t times[kNumProcCpuStates] = {0};
```
thomas-moulard:
```suggestion
  std::array<size_t, kNumProcCpuStates> times{};
```

please avoid unsafe C arrays, and prefer STL structures with bound checkings.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,89 @@
...
+    if (!stat_cpu_line.compare(0, CPU_LABEL_LENGTH, CPU_LABEL)) {
+      std::istringstream ss(stat_cpu_line);
+
+      ss >> parsed_data.cpu_label;
+
+      for (int i = 0; i < ProcCpuData::kNumProcCpuStates; ++i) {
+        ss >> parsed_data.times[i];
+      }
+
+      return parsed_data;
```
thomas-moulard: Please add error management. We should fail if `!ss.good()`. If for whatever reason we fail to parse, this will continue silently.

https://en.cppreference.com/w/cpp/io/basic_ios/good

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  return cpu_percentage;
+}
+
+ProcCpuData LinuxCpuMeasurementNode::makeSingleMeasurement()
+{
+  std::ifstream stat_file(PROC_STAT_FILE);
+  std::string line;
+  std::getline(stat_file, line);
+  stat_file.close();
```
thomas-moulard: remove, this is not needed.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Node that periodically calculates the % of active CPU by
+ * reading /proc/stat.
+ */
+class LinuxCpuMeasurementNode : public PeriodicMeasurementNode
+{
+public:
+  static constexpr const char PROC_STAT_FILE[] = "/proc/stat";
+  static constexpr const char CPU_LABEL[] = "cpu";
+  static constexpr const size_t CPU_LABEL_LENGTH = 3;
```
thomas-moulard: Does this need to be public? If no, could we move this to the cpp file?

Generally, it is good practice to avoid exposing things we don't need to.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
```diff
@@ -0,0 +1,87 @@
...
+   * Compute the percentage of CPU active.
+   *
+   * @param measurement1 the first measurement
+   * @param measurement2 the second measurement (made after the first)
+   * @return percentage of CPU active
+   */
+  static double computeCpuActivePercentage(
+    const ProcCpuData & measurement1,
+    const ProcCpuData & measurement2);
+
```
thomas-moulard: Consider moving those out of the class if they aren't unrelated. C++ is more like Python than Java, `static` functions are better avoided unless they serve a clear purpose (e.g. [named constructor idiom](http://www.cs.technion.ac.il/users/yechiel/c++-faq/named-ctor-idiom.html))

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -61,6 +84,19 @@ std::string PeriodicMeasurementNode::getStatusString() const
...
+  (publish_period_ != PeriodicMeasurementNode::DEFAULT_PUBLISH_WINDOW ?
+  std::to_string(publish_period_.count()) + "ms" : "None") <<
+    ", clear_measurements_on_publish_=" << clear_measurements_on_publish_ <<
     ", " << Collector::getStatusString();
   return ss.str();
 }
+
+void PeriodicMeasurementNode::performPeriodicMeasurement()
+{
+  double measurement = periodicMeasurement();
```
thomas-moulard:
```suggestion
  const double measurement = periodicMeasurement();
```

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
thomas-moulard: Also, consider moving that at the top-level, using an `enum class` prevents naming clash.

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include <string>
+#include <sstream>
+
+/**
+ * Struct containing data parsed from a /proc/stat cpu line
+ */
+struct ProcCpuData
+{
+  static constexpr const char EMPTY_LABEL[] = "empty";
```
thomas-moulard: Do we need to make this public?
Consider using `std::string` instead of C strings whenever possible

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -0,0 +1,78 @@
...
+#ifndef SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
+#define SYSTEM_METRICS_COLLECTOR__CPU_DATA_HPP_
+
+#include <string>
+#include <sstream>
+
+/**
+ * Struct containing data parsed from a /proc/stat cpu line
+ */
+struct ProcCpuData
```
thomas-moulard:
```suggestion
class ProcCpuData
```

Let's avoid using struct for non-POD types. This is confusing for readers who usually expect `struct` objects to be record-like structures, not full blown classes like this one with a ctor, etc.

thomas-moulard: Consider making the ctor explicit for readability, and making this class non-copyable, unless you actually want it to be this way.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+LinuxCpuMeasurementNode::LinuxCpuMeasurementNode(
+  const std::string & name,
+  const std::chrono::milliseconds measurement_period,
+  const std::string & topic,
+  const std::chrono::milliseconds & publish_period)
+: PeriodicMeasurementNode(name, measurement_period, topic, publish_period)
+{}
+
+ProcCpuData LinuxCpuMeasurementNode::processLine(const std::string & stat_cpu_line)
```
thomas-moulard:
```suggestion
/* static */ ProcCpuData LinuxCpuMeasurementNode::processLine(const std::string & stat_cpu_line)
```

please, otherwise, it's really hard to understand why this is non-const and also don't touch `this`

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,89 @@
...
+  const double total_time = (measurement2.getIdleTime() + measurement2.getActiveTime()) -
+    (measurement1.getIdleTime() + measurement2.getActiveTime());
+
+  return 100.0 * active_time / total_time;
+}
+
+double LinuxCpuMeasurementNode::periodicMeasurement()
+{
+  const auto measurement1 = makeSingleMeasurement();
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));  // todo fixme magic constant
```
thomas-moulard:
```suggestion
  std::this_thread::sleep_for(std::chrono::milliseconds(100));  // todo fixme magic constant
```

What is the intent here? we need to fix this before merging this PR.

system_metrics_collector/test/system_metrics_collector/test_linux_cpu_measurement.cpp
```diff
@@ -0,0 +1,145 @@
...
+  bool first{true};
+};
+
+class LinuxCpuMeasurementTestFixture : public ::testing::Test
+{
+public:
+  void SetUp() override
+  {
+    const char * const argv = "d";
+    rclcpp::init(1, &argv);
```
thomas-moulard:
```suggestion
    rclcpp::init(0, nullptr);
```

This code is wrong (the same issue exists in other tests). If `rclcpp` keeps a copy of the pointer `argv`, it will segv if accessed after exiting `SetUp` scope. As this does not pass anything in particular, let's keep it safe and just set this to null.

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -0,0 +1,54 @@
...
+    "linuxCpuCollector",
+    std::chrono::milliseconds(1000),
+    "not_publishing_yet",
+    std::chrono::milliseconds(1000 * 60));
+
+  rclcpp::executors::SingleThreadedExecutor ex;
+  node->start();
+
+  // used for manual testing
+  auto r = rcutils_logging_set_logger_level(node->get_name(), RCUTILS_LOG_SEVERITY_DEBUG);
```
thomas-moulard:
```suggestion
  rcutils_logging_set_logger_level(node->get_name(), RCUTILS_LOG_SEVERITY_DEBUG);
```

Depending on the platform/flag/compiler, this will fail because `r` is an unused variable.

thomas-moulard:
This PR is in great shape for a first pass, let's chat tomorrow to see how to solve the remaining issues!

system_metrics_collector/src/moving_average_statistics/moving_average.cpp
```diff
@@ -77,13 +77,16 @@ void MovingAverageStatistics::reset()
...
-  min_ = std::min(min_, item);
-  max_ = std::max(max_, item);
-  sum_of_square_diff_from_mean_ = sum_of_square_diff_from_mean_ + (item - previous_average_) *
-    (item - average_);
+
+  if (!std::isnan(item)) {
+    count_++;
+    const double previous_average_ = average_;
+    average_ = previous_average_ + (item - previous_average_) / count_;
+    min_ = std::min(min_, item);
```
piraka9011: Something I recall from `rosbag2` for Windows support  for `std::min/max` is adding something like [this](https://github.com/ros2/rosbag2/blob/0a3ba417c017a9baa54c58e33f1d44296b8c1dec/rosbag2/CMakeLists.txt#L18) to your CMakeLists.txt

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
dabonnie: The intent was to remove the magic # and define a constant variable.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
dabonnie: The only downside here is that the static functions rely on static class constants. With this I would  like to leave in class as static functions, but open to design suggestions.

dabonnie: Resolved in person :-)

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
dabonnie: As discussed in person, let's cache the first measurement and only populated / collect upon the successive (when the measurement can actually be computed.

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
dabonnie: This was made public so it could be accessed by a unit test.

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
```diff
@@ -15,36 +15,36 @@
...
 
-#include <string>
+#include <array>
 #include <sstream>
+#include <string>
+
+/**
+ * Enum representing each item in a /proc/stat cpu line
+ */
+enum class ProcCpuStates
```
zmichaels11: Do you need to extend `int` here?
```c++
enum class ProcCpuStates : int
```

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -17,29 +17,30 @@
...
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds & publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period)
-{}
+namespace
+{
+constexpr const char PROC_STAT_FILE[] = "/proc/stat";
+constexpr const char CPU_LABEL[] = "cpu";
+constexpr const size_t CPU_LABEL_LENGTH = 3;
```
zmichaels11: you can actually use `sizeof(CPU_LABEL)` to get the length. (this includes null terminator)
https://stackoverflow.com/questions/46100310/how-to-declare-constexpr-c-string

system_metrics_collector/src/system_metrics_collector/cpu_data.hpp
dabonnie: It's initialized with a value, so I'm not sure why we need to extend. https://en.cppreference.com/w/cpp/language/enum

zmichaels11: TIL C++ well-defines `enum class` size based on the type of the first enum

PR #19, opened 2019-12-02
Remove DEFAULT_PUBLISH_WINDOW

mm318:
Remove the `DEFAULT_PUBLISH_WINDOW` sentinel value as the publishing period should not be optional anymore.

### DIFF
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
@@ -57,15 +57,15 @@ public:
    *
    * @param name the name of this node
    * @param measurement_period the period of this node, used to
-   * read measurements
+   *        read measurements
    * @param topic the topic name used for publishing
+   * @param publish_period the period at which metrics are published
    */
   LinuxCpuMeasurementNode(
     const std::string & name,
     const std::chrono::milliseconds measurement_period,
     const std::string & topic,
-    const std::chrono::milliseconds publish_period =
-    PeriodicMeasurementNode::INVALID_PUBLISH_WINDOW);
+    const std::chrono::milliseconds publish_period);
 
   virtual ~LinuxCpuMeasurementNode() = default;
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
@@ -56,10 +56,9 @@ public:
    * Construct a LinuxCpuMeasurementNode
    *
    * @param name the name of this node
-   * @param measurement_period the period of this node, used to
-   *        read measurements
+   * @param measurement_period the period of this node, used to read measurements
    * @param topic the topic name used for publishing
-   * @param publish_period the period at which metrics are published
+   * @param publish_period the period at which metrics are published. 0 ms means don't publish
    */
   LinuxCpuMeasurementNode(
     const std::string & name,
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
@@ -65,7 +65,7 @@ public:
     const std::chrono::milliseconds measurement_period,
     const std::string & topic,
     const std::chrono::milliseconds publish_period =
-    PeriodicMeasurementNode::DEFAULT_PUBLISH_WINDOW);
+    PeriodicMeasurementNode::INVALID_PUBLISH_WINDOW);
 
   virtual ~LinuxCpuMeasurementNode() = default;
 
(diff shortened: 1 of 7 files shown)


system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
```diff
@@ -65,7 +65,7 @@ class LinuxCpuMeasurementNode : public system_metrics_collector::PeriodicMeasure
     const std::chrono::milliseconds measurement_period,
```
dabonnie: Constructor doc needs to be updated with this parameter, can document behavior as well for the default.

system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
```diff
@@ -67,7 +67,7 @@ class LinuxMemoryMeasurementNode : public system_metrics_collector::PeriodicMeas
     const std::chrono::milliseconds measurement_period,
     const std::string & topic,
     const std::chrono::milliseconds publish_period =
-    PeriodicMeasurementNode::DEFAULT_PUBLISH_WINDOW);
+    PeriodicMeasurementNode::INVALID_PUBLISH_WINDOW);
```
dabonnie: Constructor doc needs to be updated with this parameter, can document behavior as well for the default.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -23,8 +23,7 @@
 namespace system_metrics_collector
 {
 
-/* static */ constexpr const std::chrono::milliseconds PeriodicMeasurementNode::
-DEFAULT_PUBLISH_WINDOW;
+constexpr const std::chrono::milliseconds PeriodicMeasurementNode::INVALID_PUBLISH_WINDOW;
```
dabonnie: Please leave the static comment.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: No longer applicable.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -54,9 +51,7 @@ bool PeriodicMeasurementNode::setupStart()
     RCLCPP_WARN(this->get_logger(), "setupStart: measurement_timer_ already exists!");
   }
 
-  if (!publish_timer_ &&
-    publish_period_ != PeriodicMeasurementNode::DEFAULT_PUBLISH_WINDOW)
-  {
+  if (!publish_timer_ && publish_period_ > std::chrono::milliseconds(0)) {
```
dabonnie: Checking in the start method is too late: this should be checked upon construction.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -88,7 +83,7 @@ std::string PeriodicMeasurementNode::getStatusString() const
     ", measurement_period=" << std::to_string(measurement_period_.count()) << "ms" <<
     ", publishing_topic=" << publishing_topic_ <<
     ", publish_period=" <<
-  (publish_period_ != PeriodicMeasurementNode::DEFAULT_PUBLISH_WINDOW ?
+  (publish_period_ > std::chrono::milliseconds(0) ?
```
dabonnie: If done in construction then this is not needed.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: The unit tests are currently deliberately passing in invalid `publish_period`s to avoid this `Collector::clearCurrentMeasurements()` call.

dabonnie: Then why not make zero valid and document as such? We only have an issue if the period is strictly less than zero.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -17,12 +17,16 @@
 
 #include <chrono>
 #include <string>
+#include <stdexcept>
```
dabonnie: Alphasort

mm318: Hmm, strange that the linters didn't catch this.

mm318: Done.

PR #21, opened 2019-12-03
[WIP] Add Windows build

thomas-moulard:


thomas-moulard:
Note: this stack contains Linux only code so this does not make sense to pursue this. Closing.

PR #16, opened 2019-11-30
Add publishing of MetricMessages to CPU/Memory measurement nodes

mm318:
Addresses https://github.com/ros-security/aws-roadmap/issues/25

The `LinuxCpuMeasurementNode` and `LinuxMemoryMeasurementNode` nodes are now capable of publishing `MetricsMessage` messages to the `system_metrics` topic:
```
# ros2 topic echo system_metrics
measurement_source_name: linuxCpuCollector
metrics_source: cpu_usage
window_start:
  sec: 1575107549
  nanosec: 274748425
window_stop:
  sec: 1575107609
  nanosec: 276182982
statistics:
- data_type: 1
  data: 30.189878463745117
- data_type: 2
  data: 24.13793182373047
- data_type: 3
  data: 52.30769348144531
- data_type: 4
  data: 5.002531051635742
- data_type: 5
  data: 58.0
---
measurement_source_name: linuxMemoryCollector
metrics_source: memory_usage
window_start:
  sec: 1575107549
  nanosec: 274973629
window_stop:
  sec: 1575107609
  nanosec: 276336083
statistics:
- data_type: 1
  data: 70.09635925292969
- data_type: 2
  data: 69.8469009399414
- data_type: 3
  data: 70.19331359863281
- data_type: 4
  data: 0.10558934509754181
- data_type: 5
  data: 59.0
---
```

Relevant unit tests were also added and are passing.

PR #22, opened 2019-12-04
Add MetricsMessagePublisher interface and implementation

mm318:
Adding the interface for `MetricsMessagePublisher` as well as some implementation of its helper functions.

The `PeriodicMeasurementNode` has been plugged in into this interface, but publishing has not been activated or made functional.

### DIFF
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -61,6 +61,11 @@ if(BUILD_TESTING)
   target_link_libraries(test_collector system_metrics_collector)
   ament_target_dependencies(test_collector rclcpp)
 
+  ament_add_gtest(test_metrics_message_publisher
+    test/system_metrics_collector/test_metrics_message_publisher.cpp)
+  target_link_libraries(test_metrics_message_publisher system_metrics_collector)
+  ament_target_dependencies(test_metrics_message_publisher metrics_statistics_msgs rclcpp)
+
   ament_add_gtest(test_linux_cpu_measurement_node
           test/system_metrics_collector/test_linux_cpu_measurement.cpp)
   target_link_libraries(test_linux_cpu_measurement_node system_metrics_collector)
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -28,21 +28,22 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
-find_package(rclcpp REQUIRED)
-find_package(rcutils REQUIRED)
 find_package(metrics_statistics_msgs REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(rcpputils REQUIRED)
+find_package(rcutils REQUIRED)
 
 add_library(system_metrics_collector SHARED
+  src/moving_average_statistics/moving_average.cpp
+  src/moving_average_statistics/types.hpp
+  src/system_metrics_collector/metrics_message_publisher.cpp
   src/system_metrics_collector/collector.cpp
   src/system_metrics_collector/linux_cpu_measurement_node.cpp
   src/system_metrics_collector/linux_memory_measurement_node.cpp
   src/system_metrics_collector/periodic_measurement_node.cpp
   src/system_metrics_collector/proc_cpu_data.cpp
-  src/moving_average_statistics/moving_average.cpp
-  src/moving_average_statistics/types.hpp
 )
-
-ament_target_dependencies(system_metrics_collector rclcpp rcpputils)
+ament_target_dependencies(system_metrics_collector rclcpp rcutils rcpputils metrics_statistics_msgs)
 ament_export_libraries(system_metrics_collector)
 
 add_executable(main src/system_metrics_collector/main.cpp)
@@ -63,22 +64,22 @@ if(BUILD_TESTING)
   ament_add_gtest(test_linux_cpu_measurement_node
           test/system_metrics_collector/test_linux_cpu_measurement.cpp)
   target_link_libraries(test_linux_cpu_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_cpu_measurement_node rclcpp)
+  ament_target_dependencies(test_linux_cpu_measurement_node rclcpp metrics_statistics_msgs)
 
   ament_add_gtest(test_linux_memory_measurement_node
           test/system_metrics_collector/test_linux_memory_measurement.cpp)
   target_link_libraries(test_linux_memory_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_memory_measurement_node rclcpp)
+  ament_target_dependencies(test_linux_memory_measurement_node rclcpp metrics_statistics_msgs)
 
   ament_add_gtest(test_moving_average_statistics
-          test/moving_average_statistics/test_moving_average_statistics.cpp)
+    test/moving_average_statistics/test_moving_average_statistics.cpp)
   target_link_libraries(test_moving_average_statistics system_metrics_collector)
   ament_target_dependencies(test_moving_average_statistics rclcpp)
 
   ament_add_gtest(test_periodic_measurement_node
     test/system_metrics_collector/test_periodic_measurement_node.cpp)
   target_link_libraries(test_periodic_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_periodic_measurement_node rclcpp)
+  ament_target_dependencies(test_periodic_measurement_node rclcpp metrics_statistics_msgs)
 
   install(TARGETS
     test_moving_average_statistics
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -43,7 +43,7 @@ add_library(system_metrics_collector SHARED
   src/system_metrics_collector/periodic_measurement_node.cpp
   src/system_metrics_collector/proc_cpu_data.cpp
 )
-ament_target_dependencies(system_metrics_collector rclcpp rcutils rcpputils metrics_statistics_msgs)
+ament_target_dependencies(system_metrics_collector metrics_statistics_msgs rclcpp rcpputils rcutils)
 ament_export_libraries(system_metrics_collector)
 
 add_executable(main src/system_metrics_collector/main.cpp)
@@ -64,22 +64,22 @@ if(BUILD_TESTING)
   ament_add_gtest(test_linux_cpu_measurement_node
           test/system_metrics_collector/test_linux_cpu_measurement.cpp)
   target_link_libraries(test_linux_cpu_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_cpu_measurement_node rclcpp metrics_statistics_msgs)
+  ament_target_dependencies(test_linux_cpu_measurement_node metrics_statistics_msgs rclcpp)
 
   ament_add_gtest(test_linux_memory_measurement_node
           test/system_metrics_collector/test_linux_memory_measurement.cpp)
   target_link_libraries(test_linux_memory_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_memory_measurement_node rclcpp metrics_statistics_msgs)
+  ament_target_dependencies(test_linux_memory_measurement_node metrics_statistics_msgs rclcpp)
 
   ament_add_gtest(test_moving_average_statistics
-    test/moving_average_statistics/test_moving_average_statistics.cpp)
+          test/moving_average_statistics/test_moving_average_statistics.cpp)
   target_link_libraries(test_moving_average_statistics system_metrics_collector)
   ament_target_dependencies(test_moving_average_statistics rclcpp)
 
   ament_add_gtest(test_periodic_measurement_node
     test/system_metrics_collector/test_periodic_measurement_node.cpp)
   target_link_libraries(test_periodic_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_periodic_measurement_node rclcpp metrics_statistics_msgs)
+  ament_target_dependencies(test_periodic_measurement_node metrics_statistics_msgs rclcpp)
 
   install(TARGETS
     test_moving_average_statistics
--- a/system_metrics_collector/src/moving_average_statistics/types.hpp
+++ b/system_metrics_collector/src/moving_average_statistics/types.hpp
@@ -15,11 +15,26 @@
 #ifndef MOVING_AVERAGE_STATISTICS__TYPES_HPP_
 #define MOVING_AVERAGE_STATISTICS__TYPES_HPP_
 
+#include <array>
+#include <cmath>
 #include <sstream>
 #include <string>
 
+#include "metrics_statistics_msgs/msg/statistic_data_type.hpp"
+
 namespace moving_average_statistics
 {
+
+/**
+ *  Enumeration of the various StatisticDataTypes
+ */
+constexpr const std::array<uint8_t, 5> STATISTICS_DATA_TYPES = {
+  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_AVERAGE,
+  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_MAXIMUM,
+  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_MINIMUM,
+  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_SAMPLE_COUNT,
+  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_STDDEV
+};
 
 /**
  *  A container for statistics data results for a set of recorded observations.
--- a/system_metrics_collector/src/moving_average_statistics/types.hpp
+++ b/system_metrics_collector/src/moving_average_statistics/types.hpp
@@ -15,26 +15,12 @@
 #ifndef MOVING_AVERAGE_STATISTICS__TYPES_HPP_
 #define MOVING_AVERAGE_STATISTICS__TYPES_HPP_
 
-#include <array>
 #include <cmath>
 #include <sstream>
 #include <string>
 
-#include "metrics_statistics_msgs/msg/statistic_data_type.hpp"
-
 namespace moving_average_statistics
 {
-
-/**
- *  Enumeration of the various StatisticDataTypes
- */
-constexpr const std::array<uint8_t, 5> STATISTICS_DATA_TYPES = {
-  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_AVERAGE,
-  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_MAXIMUM,
-  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_MINIMUM,
-  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_SAMPLE_COUNT,
-  metrics_statistics_msgs::msg::StatisticDataType::STATISTICS_DATA_TYPE_STDDEV
-};
 
 /**
  *  A container for statistics data results for a set of recorded observations.
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
@@ -117,4 +117,11 @@ system_metrics_collector::ProcCpuData LinuxCpuMeasurementNode::makeSingleMeasure
   }
 }
 
+void LinuxCpuMeasurementNode::publishStatisticMessage()
+{
+  auto msg = generateStatisticMessage(get_name(), "cpu_usage", window_start_,
+      now(), getStatisticsResults());
+  publisher_->publish(msg);
+}
+
 }  // namespace system_metrics_collector
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
@@ -25,6 +25,7 @@
 
 namespace
 {
+constexpr const char MEASUREMENT_TYPE[] = "cpu_usage";
 constexpr const char PROC_STAT_FILE[] = "/proc/stat";
 constexpr const char CPU_LABEL[] = "cpu";
 }  // namespace
@@ -119,7 +120,7 @@ system_metrics_collector::ProcCpuData LinuxCpuMeasurementNode::makeSingleMeasure
 
 void LinuxCpuMeasurementNode::publishStatisticMessage()
 {
-  auto msg = generateStatisticMessage(get_name(), "cpu_usage", window_start_,
+  auto msg = generateStatisticMessage(get_name(), MEASUREMENT_TYPE, window_start_,
       now(), getStatisticsResults());
   publisher_->publish(msg);
 }
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
@@ -25,7 +25,7 @@
 
 namespace
 {
-constexpr const char MEASUREMENT_TYPE[] = "cpu_usage";
+constexpr const char MEASUREMENT_TYPE[] = "system_cpu_usage";
 constexpr const char PROC_STAT_FILE[] = "/proc/stat";
 constexpr const char CPU_LABEL[] = "cpu";
 }  // namespace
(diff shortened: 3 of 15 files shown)


system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
```diff
@@ -0,0 +1,65 @@
...
+{
+
+/**
+ * Simple class to facilitate publishing messages containing statistics data
+ */
+class MetricsMessagePublisher
+{
+public:
+  static constexpr const std::chrono::milliseconds INVALID_PUBLISH_WINDOW =
+    std::chrono::milliseconds(0);
```
thomas-moulard: Let's avoid 0 -> -1

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @param node_name the name of the node that the data originates from
+   * @param metric_name the name of the metric ("cpu_usage", "memory_usage", etc.)
+   * @param window_start measurement window start time
+   * @param window_stop measurement window end time
+   * @param data statistics derived from the measurements made in the window
+   * @return a MetricsMessage containing the statistics in the data parameter
+   */
+  static metrics_statistics_msgs::msg::MetricsMessage generateStatisticMessage(
+    std::string node_name,
```
thomas-moulard:
```suggestion
    const std::string & node_name,
```

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
```diff
@@ -0,0 +1,65 @@
...
+    std::string metric_name,
```
thomas-moulard:
```suggestion
    const std::string & metric_name,
```

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
```diff
@@ -0,0 +1,65 @@
...
+    builtin_interfaces::msg::Time window_start,
+    builtin_interfaces::msg::Time window_stop,
+    const moving_average_statistics::StatisticData & data
+  );
+
+  /**
+   * Called via a ROS2 timer per the publish_period_. This publishes the statistics derived from
```
thomas-moulard: doc phrasing: Call...

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -24,9 +24,6 @@
 namespace system_metrics_collector
 {
 
-/* static */ constexpr const std::chrono::milliseconds PeriodicMeasurementNode::
-INVALID_PUBLISH_WINDOW;
```
thomas-moulard: revert

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -81,6 +78,10 @@ bool PeriodicMeasurementNode::setupStop()
     measurement_timer_->cancel();
     measurement_timer_.reset();
   }
+  if (publish_timer_) {
+    publish_timer_->cancel();
+    publish_timer_.reset();
+  }
```
thomas-moulard: see my previous feedback: let's throw if `setupStop()` is called but the node is already called.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -61,6 +61,8 @@ class TestPeriodicMeasurementNode : public ::system_metrics_collector::PeriodicM
     return static_cast<double>(sum.load());
   }
 
+  void publishStatisticMessage() {}
```
thomas-moulard:
```suggestion
  void publishStatisticMessage() override {}
```

(not sure)

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -109,8 +107,17 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector, publ
    * If true, then measurements are cleared after publishing data
    */
   const bool clear_measurements_on_publish_;
+
   rclcpp::TimerBase::SharedPtr measurement_timer_;
   rclcpp::TimerBase::SharedPtr publish_timer_;
+
+protected:
+  /**
+   * The tracking of the starting time of the statistics
```
thomas-moulard: Track...

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: Instead of throwing, why not simply log an error. We've followed this patter previously (file_talker, etc).

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -117,4 +117,11 @@ system_metrics_collector::ProcCpuData LinuxCpuMeasurementNode::makeSingleMeasure
   }
 }
 
+void LinuxCpuMeasurementNode::publishStatisticMessage()
+{
+  auto msg = generateStatisticMessage(get_name(), "cpu_usage", window_start_,
```
dabonnie: cpu_usage should most likely be a static const and should be system_cpu_usage.

system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.cpp
```diff
@@ -124,4 +124,11 @@ double LinuxMemoryMeasurementNode::periodicMeasurement()
   return processMemInfoLines(read_string);
 }
 
+void LinuxMemoryMeasurementNode::publishStatisticMessage()
+{
+  auto msg = generateStatisticMessage(get_name(), "memory_usage", window_start_,
```
dabonnie: Same comment as above

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
dabonnie: We might just want to define in tests because I believe it's the only place this is used.

system_metrics_collector/CMakeLists.txt
```diff
@@ -28,21 +28,22 @@ endif()
...
   src/system_metrics_collector/linux_cpu_measurement_node.cpp
   src/system_metrics_collector/linux_memory_measurement_node.cpp
   src/system_metrics_collector/periodic_measurement_node.cpp
   src/system_metrics_collector/proc_cpu_data.cpp
-  src/moving_average_statistics/moving_average.cpp
-  src/moving_average_statistics/types.hpp
 )
-
-ament_target_dependencies(system_metrics_collector rclcpp rcpputils)
+ament_target_dependencies(system_metrics_collector rclcpp rcutils rcpputils metrics_statistics_msgs)
```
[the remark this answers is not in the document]
mm318: Done.

system_metrics_collector/CMakeLists.txt
```diff
@@ -63,22 +64,22 @@ if(BUILD_TESTING)
   ament_add_gtest(test_linux_cpu_measurement_node
           test/system_metrics_collector/test_linux_cpu_measurement.cpp)
   target_link_libraries(test_linux_cpu_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_cpu_measurement_node rclcpp)
+  ament_target_dependencies(test_linux_cpu_measurement_node rclcpp metrics_statistics_msgs)
```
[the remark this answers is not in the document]
mm318: Done.

system_metrics_collector/CMakeLists.txt
```diff
@@ -63,22 +64,22 @@ if(BUILD_TESTING)
...
 
   ament_add_gtest(test_linux_memory_measurement_node
           test/system_metrics_collector/test_linux_memory_measurement.cpp)
   target_link_libraries(test_linux_memory_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_memory_measurement_node rclcpp)
+  ament_target_dependencies(test_linux_memory_measurement_node rclcpp metrics_statistics_msgs)
```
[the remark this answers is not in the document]
mm318: Done.

system_metrics_collector/CMakeLists.txt
```diff
@@ -63,22 +64,22 @@ if(BUILD_TESTING)
...
 
   ament_add_gtest(test_moving_average_statistics
-          test/moving_average_statistics/test_moving_average_statistics.cpp)
+    test/moving_average_statistics/test_moving_average_statistics.cpp)
```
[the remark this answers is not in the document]
mm318: Sorry, this was an accident. Reverted.

system_metrics_collector/CMakeLists.txt
```diff
@@ -63,22 +64,22 @@ if(BUILD_TESTING)
...
   target_link_libraries(test_moving_average_statistics system_metrics_collector)
   ament_target_dependencies(test_moving_average_statistics rclcpp)
 
   ament_add_gtest(test_periodic_measurement_node
     test/system_metrics_collector/test_periodic_measurement_node.cpp)
   target_link_libraries(test_periodic_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_periodic_measurement_node rclcpp)
+  ament_target_dependencies(test_periodic_measurement_node rclcpp metrics_statistics_msgs)
```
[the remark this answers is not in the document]
mm318: Done.

system_metrics_collector/src/moving_average_statistics/types.hpp
```diff
@@ -15,12 +15,27 @@
...
 
+#include "metrics_statistics_msgs/msg/statistic_data_type.hpp"
+
 namespace moving_average_statistics
 {
 
+/**
+ *  Enumeration of the various StatisticDataTypes
+ */
+constexpr const std::array<uint8_t, 5> STATISTICS_DATA_TYPES = {
```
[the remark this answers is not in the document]
mm318: I think I can live without this variable.

Regarding 3, the whole discussion last night was about how that won't compile. This is not a pre-C++17 static member variable.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.cpp
```diff
@@ -0,0 +1,61 @@
...
+using metrics_statistics_msgs::msg::StatisticDataType;
+
+namespace system_metrics_collector
+{
+
+/* static */
+constexpr const std::chrono::milliseconds MetricsMessagePublisher::INVALID_PUBLISH_WINDOW;
+
+MetricsMessage MetricsMessagePublisher::generateStatisticMessage(
+  std::string node_name,
```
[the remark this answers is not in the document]
mm318: Okay, I changed it.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.cpp
```diff
@@ -0,0 +1,61 @@
...
+  std::string source_name,
```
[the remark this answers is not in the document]
mm318: Okay, I changed it.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.cpp
```diff
@@ -0,0 +1,61 @@
...
+  builtin_interfaces::msg::Time window_start,
+  builtin_interfaces::msg::Time window_stop,
+  const moving_average_statistics::StatisticData & data)
+{
+  MetricsMessage msg;
+
+  msg.measurement_source_name = std::move(node_name);
+  msg.metrics_source = std::move(source_name);
```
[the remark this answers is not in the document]
mm318: Okay, I changed it.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.cpp
```diff
@@ -0,0 +1,61 @@
...
+  msg.window_start = std::move(window_start);
+  msg.window_stop = std::move(window_stop);
```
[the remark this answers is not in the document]
mm318: Okay, I changed it.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.cpp
```diff
@@ -0,0 +1,61 @@
...
+  std::string source_name,
+  builtin_interfaces::msg::Time window_start,
+  builtin_interfaces::msg::Time window_stop,
+  const moving_average_statistics::StatisticData & data)
+{
+  MetricsMessage msg;
+
+  msg.measurement_source_name = std::move(node_name);
+  msg.metrics_source = std::move(source_name);
+  msg.window_start = std::move(window_start);
```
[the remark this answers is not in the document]
mm318: Okay, I changed it.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.cpp
```diff
@@ -0,0 +1,61 @@
...
+MetricsMessage MetricsMessagePublisher::generateStatisticMessage(
+  std::string node_name,
+  std::string source_name,
+  builtin_interfaces::msg::Time window_start,
+  builtin_interfaces::msg::Time window_stop,
+  const moving_average_statistics::StatisticData & data)
+{
+  MetricsMessage msg;
+
+  msg.measurement_source_name = std::move(node_name);
```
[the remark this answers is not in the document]
mm318: Okay, I changed it.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
mm318: We're going in circles on this. Can we punt on this?

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
mm318: Okay, I changed it.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
mm318: Okay, I changed it.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: This has been moved to `metrics_message_publisher.cpp`.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
mm318: This is just there for the build to pass. It serves no functional purpose in this unit test.

I put the `override` keyword there anyway.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp

system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
thomas-moulard: No, all review comments need to be resolved. Let's go with Devin's suggestions.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -19,22 +19,23 @@
 #include <chrono>
 #include <string>
 
+#include "rclcpp/rclcpp.hpp"
+#include "metrics_statistics_msgs/msg/metrics_message.hpp"
+
```
thomas-moulard: alphasort or an empty line is missing here.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -109,8 +107,17 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector, publ
...
   rclcpp::TimerBase::SharedPtr measurement_timer_;
   rclcpp::TimerBase::SharedPtr publish_timer_;
+
+protected:
+  /**
+   * Track the starting time of the statistics
+   */
+  rclcpp::Time window_start_;
+
+  rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr publisher_;
```
thomas-moulard: The order for classes is public -> protected -> private (not public -> private -> protected like here).
See https://google.github.io/styleguide/cppguide.html#Declaration_Order

> A class definition should usually start with a public: section, followed by protected:, then private:. Omit sections that would be empty.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
mm318: Okay, then I believe I should remove all the `publish_timer_` along with this constant, since it's not functional right now anyway. I'll put it back in a future PR.

dabonnie: > Okay, then I believe I should remove all the `publish_timer_` along with this constant, since it's not functional right now anyway. I'll put it back in a future PR.

As long as we can continue with working unit tests, as the tests rely on an option to disable publishing.

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -25,6 +25,8 @@
 #include "../../src/system_metrics_collector/linux_cpu_measurement_node.hpp"
 #include "../../src/system_metrics_collector/linux_memory_measurement_node.hpp"
 
+static constexpr const char STATISTICS_TOPIC_NAME[] = "system_metrics";
```
zmichaels11: Nit: You used anonymous namespaces elsewhere, for consistency this should also be in an anonymous namespace.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.cpp
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
dabonnie: Can you please add a unit test for this class? I wouldn't expect a full mock, but you can just extend the publish method and assert that the message fields are created correctly.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -90,9 +91,6 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector, publ
    */
   bool setupStop() override;
 
-  // todo implement on publish timer callback, check if we need to clear the window
```
dabonnie: Was this implemented? If so I think I missed where.

mm318: This was implemented in https://github.com/ros-tooling/system_metrics_collector/pull/16. I'll put this comment back for this PR.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: I'll just assert as @thomas-moulard had suggested in https://github.com/ros-tooling/system_metrics_collector/pull/16.

The reason for that is we don't want `setupStart()` and `setupStop()` to partially initialize/de-initialize, for example set/cancel the `measurement_timer_` but not the `publish_timer_`.

None of this could really happen anyway as `setupStart()`/`setupStop()` cannot be called by anything other than `Collector::start()`/`Collector::stop()`. Logging an error would be too silent.

dabonnie: I don't agree with logic. Stop was meant to be treated as a teardown function, so throwing an exception if start was never called is undesirable. Logging in this case is appropriate.

If we want to check for this case if started, then we should add an atomic started_ member to the PeriodicMeasurementNode.

mm318: The logic is it makes more sense for log message to be in `Collector::stop()`, because that's something that the user can actually call. `PeriodicMeasurementNode::setupStop()` is already guarded by the logic in `Collector::stop()` from being called if the node has not already started.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
mm318: Did alphasort.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/main.cpp

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.cpp
mm318: Added unit test.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
mm318: I've put this comment back.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: > The logic is it makes more sense for log message to be in `Collector::stop()`, because that's something that the user can actually call. `PeriodicMeasurementNode::setupStop()` is already guarded by the logic in `Collector::stop()` from being called if the node has not already started.

Then we need the relevant checks there as well, otherwise setupStop will throw in an "ok" case.

> If `PeriodicMeasurementNode::setupStop()` were called incorrectly, that implies a logic error or bug in our code, which warrants more than a log message.

Again, I don't agree - there's no logic error in attempting to tear down resources that weren't created. If resources were started / created in the constructor then I would agree.

Re-reading @thomas-moulard 's post above, to me he means throw if stop was already called. This would warrant the atomic state variable I mentioned previously. My fear is without using the state variable the handling of state based solely on member initialization state is complicated, dangerous, and error prone.

dabonnie:
Thanks for adding the test. We should add a test for the publish mechanism as well (add test in the periodic measurement node test file) - either here or in another PR.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: `setupStop()` is not supposed to be called if `setupStart()` has not been previously called. Full stop. Regardless if resources are going to be torn down or not.

mm318: And the only way `setupStop()` is going to be called is via the implementation in `Collector`. And `Collector` does have this "atomic" state variable that you want (well, it's a `bool` and `mutex` pair, as opposed to a `atomic<bool>`).

dabonnie: Yup, I'm talking with @thomas-moulard and had forgot about that. So if we assert in setupStart / setupStop with the correct state then it's fine to throw.

mm318: So what more changes are needed?

dabonnie: > So what more changes are needed?

Since setupStart is only called if the Collector has been started, we can assert that the publishers, timers, etc are non-null while throwing otherwise. 

Does this make sense? It's the conclusion that @thomas-moulard and I came to during a hallway discussion. If it isn't clear let's chat on a call, as this is a fairly lengthy thread (my bad).

mm318: Do you mean `assert(...)` or `if (...) throw ...`? It doesn't make sense to have both.

This PR is already doing `assert(...)`, and I think it's fine.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
```diff
@@ -0,0 +1,62 @@
...
+   * @param window_start measurement window start time
+   * @param window_stop measurement window end time
+   * @param data statistics derived from the measurements made in the window
+   * @return a MetricsMessage containing the statistics in the data parameter
+   */
+  static metrics_statistics_msgs::msg::MetricsMessage generateStatisticMessage(
+    const std::string & node_name,
+    const std::string & metric_name,
+    const builtin_interfaces::msg::Time & window_start,
+    const builtin_interfaces::msg::Time & window_stop,
```
zmichaels11:
```c++
    const builtin_interfaces::msg::Time window_start,
    const builtin_interfaces::msg::Time window_stop,
```
Cheap to copy objects should be passed by value. 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in gives a suggestion of a struct of `sizeof` 2-3 words will be cheap.

PR #23, opened 2019-12-04
Add utilities file

dabonnie:
Moved memory functions to a utilities file. Did not move any CPU functions yet as they aren't currently needed to be shared.


thomas-moulard:
After talking offline, this code is not tested yet - depending on how much code is needed for tests, either adding it as this PR is ok, or taking that as a different PR could also be possible

dabonnie:
> After talking offline, this code is not tested yet - depending on how much code is needed for tests, either adding it as this PR is ok, or taking that as a different PR could also be possible

The code moved had relevant tests, but not in a separate file. Just added changes which moved the relevant tests to a separate file for more granularity.

dabonnie:
I'm not sure why the gh actions are being cancelled. Did we possibly exceed a limit, @thomas-moulard ?

Issue #24, opened 2019-12-05
System CPU Parsing Resolution


dabonnie:
When parsing values from /proc/stat for the cpu line, it was observed that a parsed value overflowed an int, which caused the stringstream.good() to return false.

dabonnie:
Used proper bounds here: https://github.com/ros-tooling/system_metrics_collector/pull/29

PR #25, opened 2019-12-05
Fix parsing overflow issue (#24)

dabonnie:
* Fix parsing overflow issue (#24)
* Update system cpu measurement name to reflect units

### DIFF
--- a/system_metrics_collector/test/system_metrics_collector/test_linux_cpu_measurement.cpp
+++ b/system_metrics_collector/test/system_metrics_collector/test_linux_cpu_measurement.cpp
@@ -34,6 +34,8 @@ constexpr const char proc_sample_2[] =
 constexpr const std::chrono::milliseconds TEST_PERIOD =
   std::chrono::milliseconds(50);
 constexpr const double CPU_ACTIVE_PERCENTAGE = 2.7239908106334099;
+constexpr const char proc_sample_resolution_test[] =
+  "cpu  57211920 335926 18096939 2526329830 14818556 0 1072048 0 0 0\n";
 }  // namespace
 
 class TestLinuxCpuMeasurementNode : public system_metrics_collector::LinuxCpuMeasurementNode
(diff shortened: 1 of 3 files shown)


system_metrics_collector/test/system_metrics_collector/test_linux_cpu_measurement.cpp
```diff
@@ -34,6 +34,8 @@ constexpr const char proc_sample_2[] =
 constexpr const std::chrono::milliseconds TEST_PERIOD =
   std::chrono::milliseconds(50);
 constexpr const double CPU_ACTIVE_PERCENTAGE = 2.7239908106334099;
+constexpr const char proc_sample_resolution_test[] =
```
mm318: Why not put this (and other `proc_sample`s) into `test_constants.hpp`?

dabonnie: > Why not put this (and other `proc_sample`s) into `test_constants.hpp`?

I created the test constants file for the shared constants between tests, but didn't have a strong opinion about the location of the constants in question. For consistency I think it makes sense to move them so will do so - thanks for the suggestion.

mm318: I did suggest it for consistency.

However, if `test_constants.hpp` were a clean slate, my first thought would be to keep the memory strings in `test_linux_memory_measurement.cpp` and the CPU strings kept here in `test_linux_cpu_measurement.cpp`, but things like `TEST_PERIOD` and `INVALID_PUBLISH_WINDOW` be moved to `test_constants.hpp`.

With my upcoming unit tests, there will be more things to put into `test_constants.hpp`, so don't worry about it being under-utilized.

emersonknapp: seems to me like it'd make sense to just put this in the test itself, until it's actually being shared...?

mm318: `TEST_PERIOD` and `INVALID_PUBLISH_WINDOW` are examples of what's already common between `test_linux_memory_measurement.cpp` and `test_linux_cpu_measurement.cpp`.

dabonnie: FYI the constants defined in the ```test_constants.hpp``` file are shared. Since none of the CPU constants are currently shared I'll forgo moving them and merge this PR.

PR #31, opened 2019-12-10
Remove clear_measurements_on_publish

mm318:
As alluded to at https://github.com/ros-tooling/system_metrics_collector/pull/16#discussion_r352762504, this change removes the `clear_measurements_on_publish` argument from `PeriodicMeasurementNode`, treating it as if it was always `true`.

### DIFF
--- a/system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
+++ b/system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
@@ -45,11 +45,8 @@ public:
     const std::string & name,
     const std::chrono::milliseconds measurement_period,
     const std::string & publishing_topic,
-    const std::chrono::milliseconds publish_period,
-    const bool clear_measurements_on_publish)
-  : PeriodicMeasurementNode(name, measurement_period, publishing_topic, publish_period,
-      clear_measurements_on_publish)
-  {}
+    const std::chrono::milliseconds publish_period)
+  : PeriodicMeasurementNode(name, measurement_period, publishing_topic, publish_period) {}
   virtual ~TestPeriodicMeasurementNode() = default;
 
   int getNumPublished() const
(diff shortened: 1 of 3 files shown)


system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -120,14 +117,17 @@ class PeriodicMeasurementTestFixure : public ::testing::Test
   }
 
 protected:
+  static constexpr std::chrono::milliseconds INFREQUENT_PUBLISH_PERIOD = 2 *
+    test_constants::TEST_DURATION;
```
thomas-moulard: adding a comment to explain why this value is OK would be helpful, not sure if I understand very well what all those constants mean, and why this is safe and this test won't be flaky.

thomas-moulard: also, it's unclear what's the value here there. Once every 10ms, 10 seconds? INFREQUENT is not really helpful in this context.

mm318: @thomas-moulard and @dabonnie, can you take a look again?

thomas-moulard: Thanks @mm318 - before merging could you run this with `--gtest_repeat=100000` to make sure it's not flaky?

mm318: I was able to repeat the test 14000+ times without fail before running into some unrelated RTPS error (like this https://github.com/aliasrobotics/aztarna/issues/41)...

I am confident that the test is not flaky, as we should absolutely not see any publishing if the publishing period is set to *twice* that of the entire allotted test duration, even at the millisecond scale.

If we do see some publishing occur given this much margin (of the entire test duration), I would be concerned about the ROS2 executor.

PR #34, opened 2019-12-10
Install system metrics collector lib and executable

ahcorde:
Install system metrics collector lib and executable

### DIFF
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -106,4 +106,16 @@ if(BUILD_TESTING)
     lib/${PROJECT_NAME})
 endif()
 
+install(TARGETS
+  main
+  DESTINATION
+  lib/${PROJECT_NAME}
+)
+
+install(TARGETS ${PROJECT_NAME}
+  DESTINATION lib
+)
+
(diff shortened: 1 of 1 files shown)


dabonnie:
Hey @ahcorde.

Thanks for the PR! However, since we're currently in active development we'd prefer to have our headers private until we finalize our API. We can review this in next week's ROS2 Tooling Working Group and, if you're curious, you can see our team roadmap here: https://app.zenhub.com/workspaces/aws-robotics-open-source-roadmap-5d9662bf4e1ec400011c53c6/board?repos=212657172

thomas-moulard:
I think the part which is confusing is that the package is providing a lib, it should not, it should be a single binary, and this binary should be installed IMHO.

zmichaels11:
@ahcorde 
Can you close this PR since this is work we do not plan to include yet?
I opened an issue #35 for this work, which we're gating after completion of https://github.com/ros-security/aws-roadmap/issues/3

ahcorde:
Hi everyone!

Done, I pushed only the installation.

I was trying to use you library in an external executable. I will be watching you repository for futures updates. Let me know if can help with something

thomas-moulard:
Merged, thanks a lot for the quick turn around on this PR!

PR #37, opened 2019-12-11
Fix function erroneously declared as static

thomas-moulard:
Functions declared in headers should never be static.
In the previous implementation, statisticsDataToString was duplicated
in each compilation unit as the function was both defined in the header
and declared as static.

### DIFF
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -35,6 +35,7 @@ find_package(rcutils REQUIRED)
 
 add_library(system_metrics_collector SHARED
   src/moving_average_statistics/moving_average.cpp
+  src/moving_average_statistics/types.cpp
   src/moving_average_statistics/types.hpp
   src/system_metrics_collector/metrics_message_publisher.cpp
   src/system_metrics_collector/collector.cpp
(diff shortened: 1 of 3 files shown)


mm318:
How about make it a member function of `StatisticData`?

thomas-moulard:
It's generally better to avoid fixing bugs and refactoring in the same PR. I don't have a strong opinion one way or another (except that a header called `types.hpp` should not have functions defined). My goal here was to try to get rid of the partial hit on the function. I thought it was because of the issue with static but it seems I was wrong. We can still merge this I guess as this is a step in the right direction (?)

system_metrics_collector/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ find_package(rcutils REQUIRED)
 
 add_library(system_metrics_collector SHARED
   src/moving_average_statistics/moving_average.cpp
+  src/moving_average_statistics/types.cpp
   src/moving_average_statistics/types.hpp
```
mm318: The `.hpp` file should not be here.

thomas-moulard: https://stackoverflow.com/questions/36174499/why-add-header-files-into-add-library-add-executable-command-in-cmake

Some people prefer that style because then you get the headers in the auto-generated IDE solutions.

thomas-moulard:
I don't understand why the partial hit is still there - this is why the coverage tests is failing. Merging as I don't know how long it will take to understand the root cause here.

system_metrics_collector/CMakeLists.txt
mm318: I don't think that's the intention here, otherwise this is a consistency issue, because why would this be the only header file under `add_library()`?

PR #46, opened 2019-12-19
Add new package to view system metric data

dabonnie:
This PR adds the files generated by ros2 pkg create for a python project.

### DIFF
--- a/system_metrics_collector/src/system_metrics_collector/main.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/main.cpp
@@ -59,7 +59,7 @@ int main(int argc, char ** argv)
     std::make_shared<system_metrics_collector::LinuxProcessMemoryMeasurementNode>(
     "linuxProcessMemoryCollector",
     1000ms,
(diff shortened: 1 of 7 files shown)


system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -59,7 +59,7 @@ int main(int argc, char ** argv)
     std::make_shared<system_metrics_collector::LinuxProcessMemoryMeasurementNode>(
     "linuxProcessMemoryCollector",
     1000ms,
-    "not_publishing_yet",
+    STATISTICS_TOPIC_NAME,
```
mm318: This system metric viewer will receive process metrics as well and filter them out?

dabonnie: That's my current plan, yes. The viewing problem can be split into the following:

1). Logging the data into a parseable format for a display tool
2). Parse the logged data
3). Plot the data

PR #51, opened 2019-12-20
Style guide constants

dabonnie:
* Update style guide constants to follow ROS2 (Google) naming convention
* Use curly initializers where applicable


system_metrics_collector/src/system_metrics_collector/utilities.cpp
```diff
@@ -63,7 +63,7 @@ ProcCpuData processStatCpuLine(const std::string & stat_cpu_line)
   ProcCpuData parsed_data;
 
   if (!stat_cpu_line.empty()) {
-    if (!stat_cpu_line.compare(0, strlen(CPU_LABEL), CPU_LABEL)) {
+    if (!stat_cpu_line.compare(0, strlen(kCpuLabel), kCpuLabel)) {
```
zmichaels11: In another PR, can you insert a comment here as to what this line is doing?

dabonnie: Sure!

Issue #50, opened 2019-12-20
Follow ROS2 Style Guide


dabonnie:
## Description

Many constants defined in this project do not follow the guidelines for [ROS2 Development](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#id1). These should be respected.

## Expected Behavior
Constants following style guide naming convention.

## Actual Behavior
Constant naming does not follow the style guide naming convention.

dabonnie:
Opened https://github.com/ros-tooling/system_metrics_collector/pull/51

mm318:
There's more to this than just the names of the constants.

thomas-moulard:
@mm318 Please list the necessary changes to make your comment actionable.

mm318:
The other issues that are immediately noticeable are:
- Function names need to `UpperCamelCase`
- Class member variables need to be suffixed with `_`
These are being addressed in https://github.com/ros-tooling/system_metrics_collector/pull/53.

mm318:
Also found that some class and function comments start with a noun (a subject) or with a singular verb (imperative mood) and do not end with proper punctuation (a period).

Looking at the [examples](https://google.github.io/styleguide/cppguide.html#Class_Comments), it seems like both class and function comments should be in the indicative mood without a subject (the subject is implied to be the class or function that the comment is for) with proper punctuation.

dabonnie:
Closed via https://github.com/ros-tooling/system_metrics_collector/pull/51 and https://github.com/ros-tooling/system_metrics_collector/pull/56

PR #53, opened 2019-12-21
Follow function naming conventions of ROS2 style guide

mm318:
This is a continuation of https://github.com/ros-tooling/system_metrics_collector/pull/51 and should complete https://github.com/ros-tooling/system_metrics_collector/issues/50.

### DIFF
--- a/system_metrics_collector/src/system_metrics_collector/collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.cpp
@@ -26,7 +26,7 @@ namespace system_metrics_collector
 
 bool Collector::Start()
 {
-  std::unique_lock<std::mutex> ulock(mutex_);
+  std::unique_lock<std::mutex> ulock(mutex);
   if (started_) {
     return false;
   }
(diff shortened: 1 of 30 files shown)


thomas-moulard:
This LGTM, this needs to be merged into master. There is no need to chain PRs for independent changes. You can just send 10 diffs against master. If there are minor conflicts, they can be resolved independently post review.

system_metrics_collector/src/system_metrics_collector/collector.cpp
```diff
@@ -24,58 +24,58 @@
...
+  return SetupStart();
 }
 
-bool Collector::stop()
+bool Collector::Stop()
 {
   bool ret = false;
   {
-    std::unique_lock<std::mutex> ulock(mutex);
+    std::unique_lock<std::mutex> ulock(mutex_);
```
zmichaels11:
```c++
std::unique_lock<std::mutex> ulock{mutex_};
```

mm318: Yes, it does. Switched over to using curly brace initialization.

PR #59, opened 2020-01-02
Add process CPU usage measurement node

mm318:
This implements the node that does per-process CPU usage collection.

Example debug output:
```
$ ./install/system_metrics_collector/lib/system_metrics_collector/main | grep linuxProcessCpuCollector
[DEBUG] [1578005328.012000628] [linuxProcessCpuCollector]: PerformPeriodicMeasurement: nan
[DEBUG] [1578005328.012084643] [linuxProcessCpuCollector]: name=linuxProcessCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, started=true, avg=nan, min=nan, max=nan, std_dev=nan, count=0
[DEBUG] [1578005329.011345403] [linuxProcessCpuCollector]: PerformPeriodicMeasurement: 3.670513
[DEBUG] [1578005329.011465002] [linuxProcessCpuCollector]: name=linuxProcessCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, started=true, avg=3.670513, min=3.670513, max=3.670513, std_dev=0.000000, count=1
[DEBUG] [1578005330.011254244] [linuxProcessCpuCollector]: PerformPeriodicMeasurement: 3.843482
[DEBUG] [1578005330.011410743] [linuxProcessCpuCollector]: name=linuxProcessCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, started=true, avg=3.756997, min=3.670513, max=3.843482, std_dev=0.086485, count=2
[DEBUG] [1578005331.011297740] [linuxProcessCpuCollector]: PerformPeriodicMeasurement: 4.094694
[DEBUG] [1578005331.011433186] [linuxProcessCpuCollector]: name=linuxProcessCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, started=true, avg=3.869563, min=3.670513, max=4.094694, std_dev=0.174150, count=3
[DEBUG] [1578005332.025331858] [linuxProcessCpuCollector]: PerformPeriodicMeasurement: 4.664516
[DEBUG] [1578005332.025468622] [linuxProcessCpuCollector]: name=linuxProcessCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, started=true, avg=4.068301, min=3.670513, max=4.664516, std_dev=0.375815, count=4
[DEBUG] [1578005333.028630184] [linuxProcessCpuCollector]: PerformPeriodicMeasurement: 3.307438
[DEBUG] [1578005333.028917962] [linuxProcessCpuCollector]: name=linuxProcessCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, started=true, avg=3.916129, min=3.307438, max=4.664516, std_dev=0.453449, count=5
[DEBUG] [1578005334.011455014] [linuxProcessCpuCollector]: PerformPeriodicMeasurement: 3.533342
[DEBUG] [1578005334.011652595] [linuxProcessCpuCollector]: name=linuxProcessCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, started=true, avg=3.852331, min=3.307438, max=4.664516, std_dev=0.437832, count=6
[DEBUG] [1578005335.012703392] [linuxProcessCpuCollector]: PerformPeriodicMeasurement: 5.471830
[DEBUG] [1578005335.012868267] [linuxProcessCpuCollector]: name=linuxProcessCpuCollector, measurement_period=1000ms, publishing_topic=not_publishing_yet, publish_period=60000ms, started=true, avg=4.083688, min=3.307438, max=5.471830, std_dev=0.696755, count=7
```

Example topic output:
```
$ ros2 topic echo /not_publishing_yet
measurement_source_name: linuxProcessCpuCollector
metrics_source: 23296_cpu_percent_used
window_start:
  sec: 1578005387
  nanosec: 13896897
window_stop:
  sec: 1578005447
  nanosec: 16203847
statistics:
- data_type: 1
  data: 3.9269039630889893
- data_type: 3
  data: 10.672758102416992
- data_type: 2
  data: 2.0373220443725586
- data_type: 5
  data: 60.0
- data_type: 4
  data: 1.0283935070037842
---
.
.
.
---
measurement_source_name: linuxProcessCpuCollector
metrics_source: 23296_cpu_percent_used
window_start:
  sec: 1578005447
  nanosec: 16265587
window_stop:
  sec: 1578005507
  nanosec: 18301838
statistics:
- data_type: 1
  data: 3.6470861434936523
- data_type: 3
  data: 8.337632179260254
- data_type: 2
  data: 1.4981666803359985
- data_type: 5
  data: 60.0
- data_type: 4
  data: 1.2943096160888672
---
```

### DIFF
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
@@ -57,13 +57,18 @@ bool LinuxProcessCpuMeasurementNode::SetupStart()
 
 double LinuxProcessCpuMeasurementNode::PeriodicMeasurement()
 {
-  const auto current_measurement = MeasurePidCpuTime();
+  const auto current_measurement = MakeSingleMeasurement();
 
   const auto cpu_percentage = ComputePidCpuActivePercentage(last_measurement_, current_measurement);
 
   last_measurement_ = current_measurement;
 
   return cpu_percentage;
+}
+
+ProcPidCpuData LinuxProcessCpuMeasurementNode::MakeSingleMeasurement()
+{
+  return MeasurePidCpuTime();
 }
 
 std::string LinuxProcessCpuMeasurementNode::GetMetricName() const
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
@@ -1,0 +1,74 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "linux_process_cpu_measurement_node.hpp"
+
+#include <sys/sysinfo.h>
+#include <sys/types.h>
+
+#include <chrono>
+#include <cmath>
+#include <fstream>
+#include <sstream>
+#include <string>
+#include <tuple>
+
+#include "rcutils/logging_macros.h"
+
+namespace
+{
+
+constexpr const char kProcStatFile[] = "/proc/stat";
+constexpr const char kProc[] = "/proc/";
+constexpr const char kStat[] = "/stat";
+constexpr const char kMetricName[] = "_cpu_percent_used";
+
+}  // namespace
+
+namespace system_metrics_collector
+{
+
+LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
+  const std::string & name,
+  const std::chrono::milliseconds measurement_period,
+  const std::string & topic,
+  const std::chrono::milliseconds publish_period)
+: PeriodicMeasurementNode(name, measurement_period, topic, publish_period),
+  metric_name_(std::to_string(GetPid()) + kMetricName)
+{
+}
+
+bool LinuxProcessCpuMeasurementNode::SetupStart()
+{
+  last_measurement_ = ProcPidCpuData();
+  return PeriodicMeasurementNode::SetupStart();
+}
+
+double LinuxProcessCpuMeasurementNode::PeriodicMeasurement()
+{
+  const ProcPidCpuData current_measurement = MeasurePidCpuTime();
+
+  const auto cpu_percentage = ComputePidCpuActivePercentage(last_measurement_, current_measurement);
+
+  last_measurement_ = current_measurement;
+
+  return cpu_percentage;
+}
+
+std::string LinuxProcessCpuMeasurementNode::GetMetricName() const
+{
+  return metric_name_;
+}
+
+}   // namespace system_metrics_collector
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
@@ -57,7 +57,7 @@ bool LinuxProcessCpuMeasurementNode::SetupStart()
 
 double LinuxProcessCpuMeasurementNode::PeriodicMeasurement()
 {
-  const ProcPidCpuData current_measurement = MeasurePidCpuTime();
+  const auto current_measurement = MeasurePidCpuTime();
 
   const auto cpu_percentage = ComputePidCpuActivePercentage(last_measurement_, current_measurement);
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
@@ -1,0 +1,94 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SYSTEM_METRICS_COLLECTOR__LINUX_PROCESS_CPU_MEASUREMENT_NODE_HPP_
+#define SYSTEM_METRICS_COLLECTOR__LINUX_PROCESS_CPU_MEASUREMENT_NODE_HPP_
+
+#include <chrono>
+#include <cmath>
+#include <string>
+#include <tuple>
+
+#include "../../src/system_metrics_collector/periodic_measurement_node.hpp"
+#include "../../src/system_metrics_collector/proc_cpu_data.hpp"
+#include "../../src/system_metrics_collector/utilities.hpp"
+
+#include "rclcpp/rclcpp.hpp"
+#include "rcutils/logging_macros.h"
+
+
+namespace system_metrics_collector
+{
+
+/**
+ * Measures the CPU percentage used by the process.
+ */
+class LinuxProcessCpuMeasurementNode : public PeriodicMeasurementNode
+{
+public:
+  /**
+   * Constructs a LinuxProcessCpuMeasurementNode.
+   *
+   * @param name the name of this node
+   * @param measurement_period the period of this node, used to read measurements
+   * @param topic the topic name used for publishing
+   * @param publish_period the period at which metrics are published.
+   */
+  LinuxProcessCpuMeasurementNode(
+    const std::string & name,
+    const std::chrono::milliseconds measurement_period,
+    const std::string & topic,
+    const std::chrono::milliseconds publish_period);
+
+protected:
+  /**
+   * Creates ROS2 timers and a publisher for periodically triggering measurements and publishing
+   * MetricsMessages.
+   * This function needs to be overridden because this class needs to also reset last_measurement_
+   * on every start or restart in addition to what is done by PeriodicMeasurementNode.
+   *
+   * @return if setup was successful
+   */
+  bool SetupStart() override;
+
+  /**
+   * Performs a periodic measurement and calculation of the percentage of CPU this process used.
+   * This obtains measurements from clock_gettime() to obtain process-specific and system-wide CPU
+   * used.
+   *
+   * @return percentage of CPU this process used
+   */
+  double PeriodicMeasurement() override;
+
+  /**
+   * Returns the name to use for this metric.
+   *
+   * @return a string of the name for this measured metric
+   */
+  std::string GetMetricName() const override;
+
+private:
+  /**
+   * The pid of this process.
+   */
+  const std::string metric_name_;
+  /**
+   * The cached processa and system measurements used in order to perform the CPU active percentage.
+   */
+  ProcPidCpuData last_measurement_;
+};
+
+}  // namespace system_metrics_collector
+
+#endif  // SYSTEM_METRICS_COLLECTOR__LINUX_PROCESS_CPU_MEASUREMENT_NODE_HPP_
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
@@ -91,7 +91,7 @@ private:
    */
   const std::string metric_name_;
   /**
-   * The cached processa and system measurements used in order to perform the CPU active percentage.
+   * The cached process and system measurements used in order to perform the CPU active percentage.
    */
   ProcPidCpuData last_measurement_;
 };
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
@@ -20,9 +20,9 @@
 #include <string>
 #include <tuple>
 
-#include "../../src/system_metrics_collector/periodic_measurement_node.hpp"
-#include "../../src/system_metrics_collector/proc_cpu_data.hpp"
-#include "../../src/system_metrics_collector/utilities.hpp"
+#include "periodic_measurement_node.hpp"
+#include "proc_cpu_data.hpp"
+#include "utilities.hpp"
 
 #include "rclcpp/rclcpp.hpp"
 #include "rcutils/logging_macros.h"
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
@@ -64,8 +64,8 @@ protected:
 
   /**
    * Performs a periodic measurement and calculation of the percentage of CPU this process used.
-   * This obtains measurements from clock_gettime() to obtain process-specific and system-wide CPU
-   * used.
+   * This obtains measurements from MakeSingleMeasurement() to obtain process-specific
+   * and system-wide CPU used.
    *
    * @return percentage of CPU this process used
    */
@@ -79,6 +79,13 @@ protected:
   std::string GetMetricName() const override;
 
 private:
+  /**
+   * Performs a single measurement of CPU data by using clock_gettime().
+   *
+   * @return ProcCpuData the measurement made
+   */
+  virtual system_metrics_collector::ProcPidCpuData MakeSingleMeasurement();
+
   /**
    * The pid of this process.
    */
--- a/system_metrics_collector/src/system_metrics_collector/main.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/main.cpp
@@ -24,6 +24,7 @@
 
 #include "../../src/system_metrics_collector/linux_cpu_measurement_node.hpp"
 #include "../../src/system_metrics_collector/linux_memory_measurement_node.hpp"
+#include "../../src/system_metrics_collector/linux_process_cpu_measurement_node.hpp"
 #include "../../src/system_metrics_collector/linux_process_memory_measurement_node.hpp"
 
 namespace
@@ -57,6 +58,13 @@ int main(int argc, char ** argv)
     kStatisticsTopicName,
     kDefaultPublishPeriod);
 
+  const auto process_cpu_node =
+    std::make_shared<system_metrics_collector::LinuxProcessCpuMeasurementNode>(
+    "linuxProcessCpuCollector",
+    kDefaultCollectPeriod,
+    "not_publishing_yet",
+    kDefaultPublishPeriod);
+
   const auto process_mem_node =
     std::make_shared<system_metrics_collector::LinuxProcessMemoryMeasurementNode>(
     "linuxProcessMemoryCollector",
@@ -67,6 +75,7 @@ int main(int argc, char ** argv)
   rclcpp::executors::MultiThreadedExecutor ex;
   cpu_node->Start();
   mem_node->Start();
+  process_cpu_node->Start();
   process_mem_node->Start();
 
   {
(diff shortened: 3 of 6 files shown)


system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -57,6 +58,13 @@ int main(int argc, char ** argv)
     kStatisticsTopicName,
     kDefaultPublishPeriod);
 
+  const auto process_cpu_node =
+    std::make_shared<system_metrics_collector::LinuxProcessCpuMeasurementNode>(
+    "linuxProcessCpuCollector",
+    kDefaultCollectPeriod,
+    "not_publishing_yet",
```
dabonnie:
```suggestion
    kStatisticsTopicName,
```

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+bool LinuxProcessCpuMeasurementNode::SetupStart()
+{
+  last_measurement_ = ProcPidCpuData();
+  return PeriodicMeasurementNode::SetupStart();
+}
+
+double LinuxProcessCpuMeasurementNode::PeriodicMeasurement()
+{
+  const ProcPidCpuData current_measurement = MeasurePidCpuTime();
```
dabonnie: Why not an auto declaration as well?

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
```diff
@@ -0,0 +1,94 @@
...
+    const std::string & name,
+    const std::chrono::milliseconds measurement_period,
+    const std::string & topic,
+    const std::chrono::milliseconds publish_period);
+
+protected:
+  /**
+   * Creates ROS2 timers and a publisher for periodically triggering measurements and publishing
+   * MetricsMessages.
+   * This function needs to be overridden because this class needs to also reset last_measurement_
```
dabonnie: Just a note: this comment was very helpful when I wondered why ```last_measurement_``` wasn't initialized in the constructor.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,222 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <cmath>
```
dabonnie: alphasort

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,222 @@
...
+using metrics_statistics_msgs::msg::StatisticDataPoint;
+using moving_average_statistics::StatisticData;
+
+namespace
+{
+constexpr const char kTestNodeName[] = "test_measure_linux_process_cpu";
+constexpr const char kTestTopic[] = "test_process_cpu_measure_topic";
+}
+
+class TestLinuxProcessCpuMeasurementNode : public system_metrics_collector::
```
dabonnie: Maybe comment that the point of this class is to expose private / protected methods for testing.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+  int GetNumReceived() const
+  {
+    return times_received_;
+  }
+
+private:
+  void MetricsMessageCallback(const MetricsMessage & msg) const
+  {
+    ASSERT_GT(6, times_received_);
```
dabonnie: Why 6?

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,222 @@
...
+  TestLinuxProcessCpuMeasurementNode(
+    const std::string & name,
+    const std::chrono::milliseconds measurement_period,
+    const std::string & topic,
+    const std::chrono::milliseconds publish_period)
+  : LinuxProcessCpuMeasurementNode(name, measurement_period, topic, publish_period) {}
+
+  double PeriodicMeasurement() override
+  {
+    LinuxProcessCpuMeasurementNode::PeriodicMeasurement();
```
dabonnie: Doesn't this assume that the measurement works on the machine running the integration tests? Is that an assumption we can make? I would consider stubbing out the file I/O and provide fake data.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
zmichaels11: Nit:
```c++
const auto current_measurement = MeasurePidCpuTime()
```

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -85,6 +94,16 @@ int main(int argc, char ** argv)
         rcutils_get_error_string().str);
     }
   }
+  {
```
emersonknapp: I didn't review the previous one, but my rule of thumb is if you copied it 3 times, it's time to factor out into a function, e.g.
```
set_node_to_debug(mem_node, "memory");
set_node_to_debug(process_mem_node, "process memory");
...
```

However, why are we setting everything to DEBUG explicitly here? Shouldn't the externally configured ROS2 logger level be respected?

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+#ifndef SYSTEM_METRICS_COLLECTOR__LINUX_PROCESS_CPU_MEASUREMENT_NODE_HPP_
+#define SYSTEM_METRICS_COLLECTOR__LINUX_PROCESS_CPU_MEASUREMENT_NODE_HPP_
+
+#include <chrono>
+#include <cmath>
+#include <string>
+#include <tuple>
+
+#include "../../src/system_metrics_collector/periodic_measurement_node.hpp"
```
emersonknapp: Why the relative include paths? These should be `#include "system_metrics_collector/foo.hpp"`

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,222 @@
...
+#include <memory>
+#include <fstream>
+#include <string>
+#include <tuple>
+#include <unordered_map>
+
+#include "metrics_statistics_msgs/msg/metrics_message.hpp"
+#include "metrics_statistics_msgs/msg/statistic_data_type.hpp"
+
+#include "../../src/system_metrics_collector/linux_process_cpu_measurement_node.hpp"
```
emersonknapp: not use relative include paths?

system_metrics_collector/src/system_metrics_collector/main.cpp
dabonnie: Right now this main serves as a local manual tester for a linux system, in order to run each of the nodes and verify functionality. This will most likely go away once proper mains + launch files exist for the system cpu + memory and the process nodes are compose-able.

emersonknapp: "most likely go away" sounds a lot to me like "accidentally ends up being used forever" lol - should be trivial to make a function or macro to remove the copypasta

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
mm318: Okay, I guess I can start doing that here in this new file. The other `.cpp` files have been doing it like this...

dabonnie: We don't have an include directory for the first release....

emersonknapp: never too late to fix an error - sorry you're the one who got caught with it heh

emersonknapp: i could be convinced that it should be a separate PR to fix the relative include paths, as long as we make sure to actually do it. i'd personally prefer to just take care of it now that we've noticed, but it's up to you

emersonknapp: @dabonnie you can `-I` / `include_directories` / `target_include_directories` any directory. The CMakeLists should be taking care of this for the targets that need to include these headers, not the source files, this pattern is brittle because it required refactoring if the directory structure ever changes

system_metrics_collector/src/system_metrics_collector/main.cpp

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
mm318: Done.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
mm318: Done.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
mm318: Added a comment here about it.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
mm318: I think if we can't make the measurement on the machine running the integration test, then it's something wrong with the way we're measuring or something is wrong with the test machine.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp

system_metrics_collector/src/system_metrics_collector/main.cpp
mm318: Implemented the `set_node_to_debug()` function.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
mm318: I've updated the paths here to be minimally relative.

We should also open an issue tracking the to-be separate PR for fixing the relative include paths.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
mm318: This case would require additional `-I` settings.

We should open an issue tracking the to-be separate PR for fixing these relative include paths.

Let's leave fixing this case to the said separate PR.

mm318: Said issue is opened here https://github.com/ros-tooling/system_metrics_collector/issues/60.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
dabonnie: I don't agree: what happens if we need to support multiple platforms? There's no requirement that the build system is able to make these measurements, which is why many of the Linux specific functions are have been separately stubbed out and tested independently.

mm318: But this is in `test_LINUX_process_cpu_measurement_node.cpp`

dabonnie: > But this is in `test_LINUX_process_cpu_measurement_node.cpp`

Right, but the unit test makes the assumption that the machine running the test supports the linux specific measurements. I think it's important to point out none of the other Linux node tests make this assumption.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
```diff
@@ -0,0 +1,94 @@
...
+   */
+  std::string GetMetricName() const override;
+
+private:
+  /**
+   * The pid of this process.
+   */
+  const std::string metric_name_;
+  /**
+   * The cached processa and system measurements used in order to perform the CPU active percentage.
```
piraka9011: typo

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -24,6 +24,7 @@
 
 #include "../../src/system_metrics_collector/linux_cpu_measurement_node.hpp"
 #include "../../src/system_metrics_collector/linux_memory_measurement_node.hpp"
+#include "../../src/system_metrics_collector/linux_process_cpu_measurement_node.hpp"
```
piraka9011: Could these be fixed in the CMakeLists?

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
```diff
@@ -0,0 +1,94 @@
...
+#include <cmath>
+#include <string>
+#include <tuple>
+
+#include "periodic_measurement_node.hpp"
+#include "proc_cpu_data.hpp"
+#include "utilities.hpp"
+
+#include "rclcpp/rclcpp.hpp"
+#include "rcutils/logging_macros.h"
```
piraka9011: Consider using logging utilities like those in [rosbag2](https://github.com/ros2/rosbag2/blob/master/rosbag2_cpp/include/rosbag2_cpp/logging.hpp)

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -57,54 +69,42 @@ int main(int argc, char ** argv)
     kStatisticsTopicName,
     kDefaultPublishPeriod);
 
+  const auto process_cpu_node =
+    std::make_shared<system_metrics_collector::LinuxProcessCpuMeasurementNode>(
```
piraka9011: Any specific reason these are shared and not unique?

system_metrics_collector/src/system_metrics_collector/main.cpp
piraka9011: Nvm, just saw the comment below

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,241 @@
...
+using metrics_statistics_msgs::msg::StatisticDataPoint;
+using moving_average_statistics::StatisticData;
+
+namespace
+{
+constexpr const char kTestNodeName[] = "test_measure_linux_process_cpu";
+constexpr const char kTestTopic[] = "test_process_cpu_measure_topic";
+}
+
+class TestLinuxProcessCpuMeasurementNode : public system_metrics_collector::
```
piraka9011: Consider calling this `fake` and not `test`. `test` would be useful for a fixture where you use the fake (IMO).

system_metrics_collector/src/system_metrics_collector/main.cpp
dabonnie: > Any specific reason these are shared and not unique?

Due to the [executor api](http://docs.ros2.org/eloquent/api/rclcpp/classrclcpp_1_1executor_1_1Executor.html#a393a16cf14b766eeb14334233f10cda7)

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
piraka9011: I think having a linux specific test might make tests flakey (see buildfarm now :upside_down_face: ) or result in more configuration down the line in the CMakeLists to make sure the tests only run on specific OS/Architectures. 

If this should _only_ run on linux, then that should be specific in the CMakeLists or in the code with an `#ifdef` IMO.

This should probably fail when running on any other system otherwise.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -0,0 +1,241 @@
...
+    EXPECT_EQ(expected_metric_name_, msg.metrics_source);
+
+    // Check measurements.
+    // There are five types of statistics:
+    // average, maximum, minimum, standard deviation, number of samples
+    EXPECT_EQ(5, msg.statistics.size());
+
+    for (const StatisticDataPoint & stat : msg.statistics) {
+      EXPECT_GT(stat.data_type, 0);
+      if (!std::isnan(stat.data)) {
```
piraka9011: Is this check b/c the data hasn't been initialized yet?

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
piraka9011: Per hallway discussion with @dabonnie, this should not call the linux specific component and instead use fake data used in `SetUp` or ctor function.

mm318: > I think it's important to point out none of the other Linux node tests make this assumption.

And consequently some of the functions under `utilities.cpp` have no code coverage.

dabonnie: > > I think it's important to point out none of the other Linux node tests make this assumption.
> 
> And consequently some of the functions under `utilities.cpp` have no code coverage.

We can always add tests where not making system specific calls (e.g. GetPID). I need to understand the coverage, but the majority of the utilities functions have corresponding unit tests covering their different return cases (via test_utilities.cpp)

mm318: Oops sorry, not `utilities.cpp`, but `LinuxCpuMeasurementNode::MakeSingleMeasurement()`, `LinuxProcessMemoryMeasurementNode::PeriodicMeasurement()`, and `LinuxMemoryMeasurementNode::PeriodicMeasurement()`.

dabonnie: I don't get the point you're trying to make, because we can walk through the existing unit tests and go over how all of the non-specific system calls directly tested. Otherwise one could mock these classes solely to call these methods directly, but that kind of test exists for the superclass.  

However, we're getting off topic here. Let's address the tests in this PR and improve coverage, etc in other issues.

mm318: > how all of the non-specific system calls directly tested.

My point is why are we not testing system-specific calls?

Anyway yes, I'll disagree and commit™ to this PR comment.

mm318: This has been mocked out via overriding `LinuxProcessCpuMeasurementNode::MakeSingleMeasurement()`.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
mm318: Fixed.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
mm318: This should be a separate PR.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
mm318: Now I'm unsure if this comment was referring to the value of the `kTestTopic` constant or the name of the class.

Anyway, I have renamed `TestLinuxProcessCpuMeasurementNode` to `MockLinuxProcessCpuMeasurementNode`, since `mock` is a more accurate description than `fake`.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
mm318: This applies to CPU measurement nodes, because calculating CPU usage requires 2 measurements, at the beginning and end of a measurement window. The first statistic message that's published has not yet seen the end of the first measurement window, so the statistic ends up being NaN.

(Memory usage measurements is not affected by this problem.)

mm318: In real world cases, the frequency of making measurements should be significantly higher than the frequency of publishing statistic messages, so NaN statistics would not be seen, but for testing I purposely made their frequencies on the same order of magnitude.

dabonnie:
Merged given passing tests and bypassing broken codecov.

PR #67, opened 2020-01-13
Make LinuxProcessCpuMeasurementNode and LinuxProcessMemoryMeasurementNode composeable

mm318:
Usage, after a successful `colcon build`:
```
source install/setup.bash
ros2 run rclcpp_components component_container
ros2 component load /ComponentManager composition composition::Talker
ros2 component load /ComponentManager system_metrics_collector system_metrics_collector::LinuxProcessCpuMeasurementNode
ros2 component load /ComponentManager system_metrics_collector system_metrics_collector::LinuxProcessMemoryMeasurementNode
```


system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
```diff
@@ -60,6 +62,25 @@ LinuxProcessMemoryMeasurementNode::LinuxProcessMemoryMeasurementNode(
...
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(
+    "linuxProcessMemoryCollector",
+    system_metrics_collector::kDefaultCollectPeriod,
+    system_metrics_collector::kStatisticsTopicName,
+    system_metrics_collector::kDefaultPublishPeriod),
+  pid_(std::to_string(GetPid())),
+  file_to_read_(kProc + std::to_string(GetPid()) + kStatm)
+{
+  const auto r = rcutils_logging_set_logger_level(get_name(), RCUTILS_LOG_SEVERITY_DEBUG);
```
dabonnie: Setting the log to debug isn't needed. Let's allow a user to set that configuration and remove our testing code.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
```diff
@@ -49,6 +51,24 @@ LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
...
+LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(
+    "linuxProcessCpuCollector",
+    system_metrics_collector::kDefaultCollectPeriod,
+    system_metrics_collector::kStatisticsTopicName,
+    system_metrics_collector::kDefaultPublishPeriod),
+  metric_name_(std::to_string(GetPid()) + kMetricName)
+{
+  const auto r = rcutils_logging_set_logger_level(get_name(), RCUTILS_LOG_SEVERITY_DEBUG);
```
dabonnie: Setting the log to debug isn't needed. Let's allow a user to set that configuration and remove our testing code.

mm318:
Things to discuss:
- `libsystem_metrics_collector.so`, `libsystem_metrics_collector_process_cpu_component.so`, and `libsystem_metrics_collector_process_memory_component.so` will have duplicate machine code, since `libsystem_metrics_collector_process_cpu_component.so` and `libsystem_metrics_collector_process_memory_component.so` need to be independent from each other. Not sure how to reliably make them get common machine code from a shared library instead of a static library
- How to pass in node name, collection/measurement period, statistics topic name, default publishing period

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp

dabonnie:
Note: configuration of the composable nodes in split out into a separate issue: https://github.com/ros-security/aws-roadmap/issues/165

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
```diff
@@ -51,6 +52,9 @@ class LinuxProcessCpuMeasurementNode : public PeriodicMeasurementNode
     const std::string & topic,
     const std::chrono::milliseconds publish_period);
 
+  SYSTEM_METRICS_COLLECTOR_PUBLIC
```
mm318: I probably actually don't need to do this (and consequently don't need to add the `visibility_control.h` header file to this project) if we're just supporting Linux.

system_metrics_collector/CMakeLists.txt
```diff
@@ -30,29 +30,65 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(metrics_statistics_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rclcpp_components REQUIRED)
 find_package(rcpputils REQUIRED)
 find_package(rcutils REQUIRED)
 
-add_library(system_metrics_collector SHARED
+add_library(system_metrics_collector_base STATIC
```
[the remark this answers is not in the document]
mm318: Currently, if `libsystem_metrics_collector_process_cpu_component.so` depends on a `libsystem_metrics_collector_base.so`, it can't find it at runtime.

I can try setting `rpath`. Setting `rpath` to a relative path is discouraged. I'll set it to an absolute path, but that means these `.so` files can't be moved elsewhere after being installed.

system_metrics_collector/src/system_metrics_collector/default_config.hpp
```diff
@@ -0,0 +1,29 @@
...
+#define SYSTEM_METRICS_COLLECTOR__DEFAULT_CONFIG_HPP_
+
+#include <chrono>
+
+namespace system_metrics_collector
+{
+
+constexpr const char kStatisticsTopicName[] = "system_metrics";
+constexpr const std::chrono::seconds kDefaultCollectPeriod{1};
+constexpr const std::chrono::minutes kDefaultPublishPeriod{1};
```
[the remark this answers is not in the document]
zmichaels11: I found a StackOverflow on this: https://stackoverflow.com/questions/50488831/use-of-constexpr-in-header-file
It looks like an alternative would be to wrap `kStatisticsTopicName` in a function.
`kDefaultCollectPeriod` and `kDefaultPublishPeriod` are probably fine since an address is not required when either are accessed.

system_metrics_collector/src/system_metrics_collector/default_config.hpp
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
dabonnie: This file duplicates constants.hpp, with some minor namespace changes.

system_metrics_collector/CMakeLists.txt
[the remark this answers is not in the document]
dabonnie: From the design review: rename this to lib_system_metrics_collector

system_metrics_collector/src/system_metrics_collector/default_config.hpp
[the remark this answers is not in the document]
zmichaels11: More on this: https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/zu1NMS0RyrM
`constexpr const char kStringLiteral[] = ...` probably should require a .cpp file when defined in a header since it does have external linkage in C++14.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
zmichaels11: This remark should probably be in the comments

system_metrics_collector/CMakeLists.txt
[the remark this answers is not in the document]
mm318: Done.

We're going with just one monolithic shared library that provides multiple components.

system_metrics_collector/src/system_metrics_collector/default_config.hpp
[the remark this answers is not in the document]
mm318: This is no longer part of this PR.

system_metrics_collector/src/system_metrics_collector/default_config.hpp
mm318: Removed `default_config.hpp`.

mm318:
@thomas-moulard, @dabonnie, and @zmichaels11, this PR is ready for re-review. It has been implemented in the way we discussed.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
piraka9011: If the file is already there, we can keep it for now and remove down the line if its causing any problems.

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.hpp
```diff
@@ -19,8 +19,9 @@
 #include <cmath>
 #include <string>
 
-#include "../../src/system_metrics_collector/periodic_measurement_node.hpp"
-#include "../../src/system_metrics_collector/utilities.hpp"
+#include "./visibility_control.h"
```
piraka9011: Couldn't you just use ` #include "visibility_control.h"` ?

system_metrics_collector/CMakeLists.txt
```diff
@@ -48,7 +49,12 @@ add_library(system_metrics_collector SHARED
   src/system_metrics_collector/proc_pid_cpu_data.cpp
   src/system_metrics_collector/utilities.cpp
 )
-ament_target_dependencies(system_metrics_collector metrics_statistics_msgs rclcpp rcpputils rcutils)
+target_compile_definitions(system_metrics_collector PRIVATE "COMPOSITION_BUILDING_DLL")
```
piraka9011: This should be
```
target_compile_definitions(system_metrics_collector PRIVATE "SYSTEM_METRICS_COLLECTOR_BUILDING_DLL")
```

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.hpp
mm318: cpplint will complain about
```
Include the directory when naming .h files  [build/include] [4]
```

It doesn't complain like this about `.hpp` files, but I don't think this header should be a `.hpp` file.

system_metrics_collector/CMakeLists.txt
mm318: Good catch!

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
mm318: I didn't intend to make it sound potentially harmful. I think it's just useless right now since we're only on Linux/ELF where all symbols are just visible by default and we're not doing anything about symbol visibility in our CMake files right now.

dabonnie: If not needed then why are we keeping it?

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
```diff
@@ -75,3 +82,11 @@ std::string LinuxProcessCpuMeasurementNode::GetMetricName() const
 }
 
 }   // namespace system_metrics_collector
+
+
+#include "rclcpp_components/register_node_macro.hpp"
```
dabonnie: Include at the top of the file? Why here?

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
```diff
@@ -91,3 +98,11 @@ uint64_t GetProcessUsedMemory(const std::string & statm_process_file_contents)
 }
 
 }   // namespace system_metrics_collector
+
+
+#include "rclcpp_components/register_node_macro.hpp"
```
dabonnie: Include at the top of the file? Why here?

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
```diff
@@ -47,6 +47,13 @@ LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
 {
 }
 
+LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
```
dabonnie: This constructor is required by the composable API (seems like to via the examples, just confirming here).

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
```diff
@@ -50,6 +51,19 @@ class LinuxProcessCpuMeasurementNode : public PeriodicMeasurementNode
     const std::string & name,
     const rclcpp::NodeOptions & options = rclcpp::NodeOptions{});
 
+  /**
+   * Constructs a LinuxProcessCpuMeasurementNode.
+   * The node name will be "linux_process_cpu_collector" by default
+   * The following parameters may be set via the rclcpp::NodeOptions:
+   * `measurement_period`: the period of this node, used to read measurements
+   * `publish_period`: the period at which metrics are published
+   *
```
dabonnie: Nit: note that this autostarts

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.hpp
```diff
@@ -57,6 +58,20 @@ class LinuxProcessMemoryMeasurementNode : public PeriodicMeasurementNode
     const std::string & name,
     const rclcpp::NodeOptions & options = rclcpp::NodeOptions{});
 
+  /**
+   * Construct a LinuxProcessMemoryMeasurementNode
+   * The node name will be "linux_process_memory_collector" by default
+   * The following parameters may be set via the rclcpp::NodeOptions:
+   * `measurement_period`: the period of this node, used to read measurements
+   * `publish_period`: the period at which metrics are published
+   *
```
dabonnie: Nit: note the auto start behavior

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
mm318: Yep.

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
mm318: The composition demos did this. Sure, I'll move it to the top of the file.

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
mm318: I have moved it to the top of the file.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.hpp
mm318: Sure. Changes related to symbol visibility have been removed from this pull request.

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.hpp
mm318: Done.

dabonnie:
Merged as build tests pass, codecov still broken.

PR #69, opened 2020-01-14
Add input argument checking to PeriodicMeasurementNode

dabonnie:
Added input validations to the PeriodicMeasurementNode constructor, documented the valid inputs in the class, and added a unit test to verify an exception is thrown if invalid input is passed.


dabonnie:
Merging given build success. Failure is due to codecov issue.

PR #71, opened 2020-01-15
Added publish period attribute as a parameter

ahcorde:
By default the publish rate is [hardcoded to 1 minute](https://github.com/ros-tooling/system_metrics_collector/blob/master/system_metrics_collector/src/system_metrics_collector/constants.hpp#L28)

This PR allows to change this value using ROS 2 parameters

### DIFF
--- a/system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
@@ -48,11 +48,17 @@ PeriodicMeasurementNode::PeriodicMeasurementNode(
     throw std::invalid_argument{"publishing_topic cannot be empty"};
   }
 
-  if (publish_period <= std::chrono::milliseconds{0}) {
+  int publish_period_param = 0;
+  declare_parameter("publish_period");
+  if (get_parameter("publish_period", publish_period_param)) {
+    publish_period_ = std::chrono::milliseconds(publish_period_param);
+  }
+
+  if (publish_period_ <= std::chrono::milliseconds{0}) {
     throw std::invalid_argument{"publish_period cannot be negative"};
   }
 
-  if (publish_period <= measurement_period) {
+  if (publish_period_ <= measurement_period) {
     throw std::invalid_argument{
             "publish_period cannot be less than or equal to the measurement_period"};
   }
--- a/system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
@@ -49,7 +49,9 @@ PeriodicMeasurementNode::PeriodicMeasurementNode(
   }
 
   int publish_period_param = 0;
-  declare_parameter("publish_period");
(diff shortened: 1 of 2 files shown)


thomas-moulard:
Thank you for the pull request! Would you need this parameter to not be read-only?

For something like a publishing period, maybe it would be preferable to declare the parameter as read-only (see https://github.com/ros2/rcl_interfaces/blob/master/rcl_interfaces/msg/ParameterDescriptor.msg and `declare_parameter` signature). This would alleviate the possibility of bug when the parameter value is changed dynamically.

ahcorde:
Sure, 

where I should document the meaning and units of this parameter?

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -48,11 +48,19 @@ PeriodicMeasurementNode::PeriodicMeasurementNode(
     throw std::invalid_argument{"publishing_topic cannot be empty"};
   }
 
-  if (publish_period <= std::chrono::milliseconds{0}) {
+  int publish_period_param = 0;
+  rcl_interfaces::msg::ParameterDescriptor descriptor;
+  descriptor.read_only = true;
+  declare_parameter("publish_period", publish_period_.count(), descriptor);
+  if (get_parameter("publish_period", publish_period_param)) {
+    publish_period_ = std::chrono::milliseconds(publish_period_param);
```
zmichaels11:
```suggestion
    publish_period_ = std::chrono::milliseconds{publish_period_param};
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -48,11 +48,19 @@ PeriodicMeasurementNode::PeriodicMeasurementNode(
     throw std::invalid_argument{"publishing_topic cannot be empty"};
   }
 
-  if (publish_period <= std::chrono::milliseconds{0}) {
+  int publish_period_param = 0;
+  rcl_interfaces::msg::ParameterDescriptor descriptor;
+  descriptor.read_only = true;
+  declare_parameter("publish_period", publish_period_.count(), descriptor);
```
zmichaels11: Nit: It might be useful to extract `publish_period` into a constant

mm318:
We're working on making all the node constructors use `rclcpp::NodeOptions`, which will most likely supersede all the changes in this pull request.

dabonnie:
> We're working on making all the node constructors use `rclcpp::NodeOptions`, which will most likely supersede all the changes in this pull request.

Not sure if you're planning on it, but it would be great to have those changes in a separate PR :-)

thomas-moulard:
@ahcorde I think it would be good to document the parameter in the ParameterDescriptor directly.
https://github.com/ros2/rcl_interfaces/blob/master/rcl_interfaces/msg/ParameterDescriptor.msg

See the field `description`. i would also set the range.

ahcorde:
closing this PR, this other PR https://github.com/ros-tooling/system_metrics_collector/pull/73/ is more complete

PR #72, opened 2020-01-15
Refactor PeriodicMeasurementNode to inherit from LifecycleNode

dabonnie:
Inherit from LifecycleNode in order to use the existing state machine for a user to start and stop collection. This PR does the following:

- refactor PeriodicMeasurementNode to inherit from LifecycleNode
- refactor unit tests to manually call lifecycle transitions
- refactor launch entry points to automatically activate and shutdown

### DIFF
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -30,6 +30,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(metrics_statistics_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
 find_package(rcpputils REQUIRED)
 find_package(rcutils REQUIRED)
 
@@ -48,7 +49,7 @@ add_library(system_metrics_collector SHARED
   src/system_metrics_collector/proc_pid_cpu_data.cpp
   src/system_metrics_collector/utilities.cpp
 )
-ament_target_dependencies(system_metrics_collector metrics_statistics_msgs rclcpp rcpputils rcutils)
+ament_target_dependencies(system_metrics_collector metrics_statistics_msgs rclcpp rclcpp_lifecycle rcpputils rcutils)
 ament_export_libraries(system_metrics_collector)
 
 add_executable(main src/system_metrics_collector/main.cpp)
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -49,7 +49,12 @@ add_library(system_metrics_collector SHARED
   src/system_metrics_collector/proc_pid_cpu_data.cpp
   src/system_metrics_collector/utilities.cpp
 )
-ament_target_dependencies(system_metrics_collector metrics_statistics_msgs rclcpp rclcpp_lifecycle rcpputils rcutils)
+ament_target_dependencies(system_metrics_collector
+        metrics_statistics_msgs
+        rclcpp
+        rclcpp_lifecycle
+        rcpputils
+        rcutils)
 ament_export_libraries(system_metrics_collector)
 
 add_executable(main src/system_metrics_collector/main.cpp)
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -67,6 +67,7 @@ ament_target_dependencies(linux_memory_collector)
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   find_package(ament_cmake_gtest REQUIRED)
+  find_package(lifecycle_msgs REQUIRED)
 
   ament_lint_auto_find_test_dependencies()
 
@@ -88,17 +89,17 @@ if(BUILD_TESTING)
   ament_add_gtest(test_linux_memory_measurement_node
           test/system_metrics_collector/test_linux_memory_measurement.cpp)
   target_link_libraries(test_linux_memory_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_memory_measurement_node metrics_statistics_msgs rclcpp)
+  ament_target_dependencies(test_linux_memory_measurement_node lifecycle_msgs metrics_statistics_msgs rclcpp)
 
   ament_add_gtest(test_linux_process_cpu_measurement_node
           test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp)
   target_link_libraries(test_linux_process_cpu_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_process_cpu_measurement_node metrics_statistics_msgs rclcpp)
+  ament_target_dependencies(test_linux_process_cpu_measurement_node lifecycle_msgs metrics_statistics_msgs rclcpp)
 
   ament_add_gtest(test_linux_process_memory_measurement_node
           test/system_metrics_collector/test_linux_process_memory_measurement_node.cpp)
   target_link_libraries(test_linux_process_memory_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_linux_process_memory_measurement_node metrics_statistics_msgs rclcpp)
+  ament_target_dependencies(test_linux_process_memory_measurement_node lifecycle_msgs metrics_statistics_msgs rclcpp)
 
   ament_add_gtest(test_moving_average_statistics
           test/moving_average_statistics/test_moving_average_statistics.cpp)
@@ -108,7 +109,7 @@ if(BUILD_TESTING)
   ament_add_gtest(test_periodic_measurement_node
     test/system_metrics_collector/test_periodic_measurement_node.cpp)
   target_link_libraries(test_periodic_measurement_node system_metrics_collector)
-  ament_target_dependencies(test_periodic_measurement_node metrics_statistics_msgs rclcpp)
+  ament_target_dependencies(test_periodic_measurement_node lifecycle_msgs metrics_statistics_msgs rclcpp)
 
   ament_add_gtest(test_utilities
           test/system_metrics_collector/test_utilities.cpp)
--- a/system_metrics_collector/src/system_metrics_collector/main.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/main.cpp
@@ -76,10 +76,10 @@ int main(int argc, char ** argv)
   set_node_to_debug(process_cpu_node.get(), "process cpu");
   set_node_to_debug(process_mem_node.get(), "process memory");
 
-  ex.add_node(cpu_node);
-  ex.add_node(mem_node);
-  ex.add_node(process_cpu_node);
-  ex.add_node(process_mem_node);
+  ex.add_node(cpu_node->get_node_base_interface());
+  ex.add_node(mem_node->get_node_base_interface());
+  ex.add_node(process_cpu_node->get_node_base_interface());
+  ex.add_node(process_mem_node->get_node_base_interface());
   ex.spin();
 
   rclcpp::shutdown();
--- a/system_metrics_collector/src/system_metrics_collector/main.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/main.cpp
@@ -79,9 +79,9 @@ int main(int argc, char ** argv)
   set_node_to_debug(process_mem_node.get(), "process memory");
 
   ex.add_node(cpu_node->get_node_base_interface());
-//  ex.add_node(mem_node->get_node_base_interface());
-//  ex.add_node(process_cpu_node->get_node_base_interface());
-//  ex.add_node(process_mem_node->get_node_base_interface());
+  ex.add_node(mem_node->get_node_base_interface());
+  ex.add_node(process_cpu_node->get_node_base_interface());
+  ex.add_node(process_mem_node->get_node_base_interface());
   ex.spin();
 
   rclcpp::shutdown();
--- a/system_metrics_collector/src/system_metrics_collector/main.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/main.cpp
@@ -69,9 +69,14 @@ int main(int argc, char ** argv)
   cpu_node->configure();
   cpu_node->activate();
 
-  mem_node->Start();
-  process_cpu_node->Start();
-  process_mem_node->Start();
+  mem_node->configure();
+  mem_node->activate();
+
+  process_cpu_node->configure();
+  process_cpu_node->activate();
+
+  process_mem_node->configure();
+  process_mem_node->activate();
 
   set_node_to_debug(cpu_node.get(), "cpu");
   set_node_to_debug(mem_node.get(), "memory");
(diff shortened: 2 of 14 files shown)


dabonnie:
Will remove from draft after adding support for the lifecycle message types (instead of comparing expected lifecycle states to integer values, compare to the actual name).

dabonnie:
Manually tested using ```ros2 topic echo /system_metrics```

```

measurement_source_name: linuxMemoryCollector
metrics_source: system_memory_percent_used
window_start: 
  sec: 1579124931         
  nanosec: 84125089
window_stop:
  sec: 1579124991
  nanosec: 83888960
statistics:
- data_type: 1
  data: 43.977439880371094
- data_type: 3
  data: 44.28981018066406
- data_type: 2
  data: 43.68082046508789
- data_type: 5
  data: 59.0
- data_type: 4
  data: 0.10391871631145477
---
measurement_source_name: linuxCpuCollector
metrics_source: system_cpu_percent_used
window_start:
  sec: 1579124931
  nanosec: 84741476
window_stop:
  sec: 1579124991
  nanosec: 84595369
statistics:
- data_type: 1
  data: 6.064512729644775
- data_type: 3
  data: 13.94169807434082
- data_type: 2
  data: 3.7809648513793945
- data_type: 5
  data: 58.0
- data_type: 4
  data: 1.8589410781860352
---

```

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -94,10 +96,10 @@ int main(int argc, char ** argv)
...
   set_node_to_debug(process_mem_node.get(), "process memory");
 
-  ex.add_node(cpu_node);
-  ex.add_node(mem_node);
-  ex.add_node(process_cpu_node);
-  ex.add_node(process_mem_node);
+  ex.add_node(cpu_node->get_node_base_interface());
+//  ex.add_node(mem_node->get_node_base_interface());
+//  ex.add_node(process_cpu_node->get_node_base_interface());
+//  ex.add_node(process_mem_node->get_node_base_interface());
```
zmichaels11: Any reason this is commented out?

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -31,7 +31,7 @@ PeriodicMeasurementNode::PeriodicMeasurementNode(
   const std::chrono::milliseconds measurement_period,
   const std::string & publishing_topic,
   const std::chrono::milliseconds publish_period)
-: Node(name),
+: rclcpp_lifecycle::LifecycleNode(name),
```
zmichaels11: Would
```c++
rclcpp_lifecycle::LifecycleNode{name}
```
work?

mm318:
Are `LifecycleNode`s composeable? https://answers.ros.org/question/333498/composable-lifecycle-node/

dabonnie:
> Are `LifecycleNode`s composeable? https://answers.ros.org/question/333498/composable-lifecycle-node/

Reading the last part of the post, I think we are OK given that this PR manually invokes the state transitions. We could do the same for the composable process nodes in their constructor (autostart flag?).

system_metrics_collector/src/system_metrics_collector/main.cpp
dabonnie: No, will revert.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: Done

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -31,7 +31,7 @@ namespace
 {
 constexpr const char kStatisticsTopicName[] = "system_metrics";
 constexpr const std::chrono::seconds kDefaultCollectPeriod{1};
-constexpr const std::chrono::minutes kDefaultPublishPeriod{1};
+constexpr const std::chrono::seconds kDefaultPublishPeriod{10};
```
dabonnie: My mistake, need to revert this.

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -84,7 +84,9 @@ int main(int argc, char ** argv)
     kDefaultPublishPeriod);
 
   rclcpp::executors::MultiThreadedExecutor ex;
-  cpu_node->Start();
+  cpu_node->configure();
+  cpu_node->activate();
+
   mem_node->Start();
```
dabonnie: Need to also configure and activate the other nodes for consistency.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -61,6 +61,15 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector,
    */
   std::string GetStatusString() const override;
 
+  rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn on_activate(
```
dabonnie: These functions need docstrings

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -82,7 +91,8 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector,
    */
   rclcpp::Time window_start_;
 
-  rclcpp::Publisher<metrics_statistics_msgs::msg::MetricsMessage>::SharedPtr publisher_;
```
dabonnie: This member needs a docstring

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -84,6 +86,31 @@ bool PeriodicMeasurementNode::SetupStart()
   return true;
 }
 
+rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+PeriodicMeasurementNode::on_activate(const rclcpp_lifecycle::State &)
+{
+  RCLCPP_DEBUG(this->get_logger(), "on_activate");
+  Start();
+  return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
```
mm318: `Start()` returns a `bool` that should probably influence the return value of this function.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -61,6 +61,36 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector,
    */
   std::string GetStatusString() const override;
 
+  /**
+   * Implementation of the on_activate transition for this LifecycleNode. This calls
+   * the Start() method.
+   *
+   * @param state input state
```
mm318: Mention input state is not used?

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -55,6 +59,16 @@ class TestPeriodicMeasurementNode : public ::system_metrics_collector::PeriodicM
...
   }
 
+  /**
+   * Return true if the lifecycle publisher is activated, false if null or not activated.
+   *
+   * @return
+   */
+  bool IsPublisherActivated() const
+  {
+    return !(publisher_ == nullptr || !publisher_->is_activated());
```
mm318: There is a lot of negation going on here, making the logic unclear. How about:
```suggestion
    return publisher_ != nullptr && publisher_->is_activated();
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: I tried that and it affected the lifecycle of the node, resulting in an improper teardown.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
dabonnie: Sure

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
dabonnie: Sure

system_metrics_collector/CMakeLists.txt
```diff
@@ -48,7 +49,7 @@ add_library(system_metrics_collector SHARED
   src/system_metrics_collector/proc_pid_cpu_data.cpp
   src/system_metrics_collector/utilities.cpp
 )
-ament_target_dependencies(system_metrics_collector metrics_statistics_msgs rclcpp rcpputils rcutils)
+ament_target_dependencies(system_metrics_collector metrics_statistics_msgs rclcpp rclcpp_lifecycle rcpputils rcutils)
```
thomas-moulard: Consider breaking that line as this starts getting too long.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -92,6 +94,44 @@ bool PeriodicMeasurementNode::SetupStart()
   return true;
 }
 
+rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+PeriodicMeasurementNode::on_activate(const rclcpp_lifecycle::State &)
+{
+  RCLCPP_DEBUG(this->get_logger(), "on_activate");
+  auto const ret = Start();
+  return ret ? rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS :
```
thomas-moulard: Consider adding `using SUCCESS = rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCES` at the top of the file to make this more readable. I.e. there should not be two ways to do a similar thing.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -92,6 +94,44 @@ bool PeriodicMeasurementNode::SetupStart()
   return true;
 }
 
+rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn
+PeriodicMeasurementNode::on_activate(const rclcpp_lifecycle::State &)
+{
+  RCLCPP_DEBUG(this->get_logger(), "on_activate");
+  auto const ret = Start();
```
thomas-moulard: Are `Start`, `Stop`, etc. private now? I think the interface should be either the lifecycle API, or the internal API, but not both.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -129,6 +171,12 @@ void PeriodicMeasurementNode::PerformPeriodicMeasurement()
 
 void PeriodicMeasurementNode::PublishStatisticMessage()
 {
+  if (!publisher_->is_activated()) {
+    RCLCPP_WARN(get_logger(), "PublishStatisticMessage: Lifecycle publisher is currently inactive."
+      " Messages will not be published.");
+    return;
+  }
```
thomas-moulard: Could this happen? Why not throwing here? It seems like the node is broken if you end up in this situation.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -57,6 +57,46 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector,
    */
   std::string GetStatusString() const override;
 
+  /**
+   * Implementation of the on_activate transition for this LifecycleNode. This calls
```
thomas-moulard: Starts the node...

https://google.github.io/styleguide/cppguide.html#Function_Comments

See the expected style for docstrings in the link.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
thomas-moulard: edit: actually in this world, the correct way to handle this is to trigger an ErrorProcessing transition IMHO
That seems to be the intent of the design: https://design.ros2.org/articles/node_lifecycle.html

It's the best of both world, the node won't die, but it also requires the user to restart the node so it can't become "ambient noise", which is the main problem with warnings.

dabonnie: > Could this happen? Why not throwing here? It seems like the node is broken if you end up in this situation.

If our logic is broken then this could happen. When using a LifecyclePublisher, the activation / deactivation responsibility is on you (see the [talker demo](https://github.com/ros2/demos/blob/master/lifecycle/src/lifecycle_talker.cpp) for a nice example). Here we explicitly activate on Start() and deactivate on Stop().

With our tests for expected publisher state, it's worth replacing the log warning with an assert because it's a logic error.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: The downside to making Collector::Start and Collector::Stop private would be for any other uses outside of ROS2 (e.g. ROS1).

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: Will do.

system_metrics_collector/CMakeLists.txt
dabonnie: Will do.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: We need to mark Start and Stop as protected, so the PeriodicNode API can call them for the state transitions.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -85,7 +99,7 @@ class TestPeriodicMeasurementNode : public ::system_metrics_collector::PeriodicM
 };
 
 /**
- * Test fixture
+ * Test fixture to test a PeriodicMeasurementNode
```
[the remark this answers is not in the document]
dabonnie: Is it? There are two test fixtures and I wanted to clarify what each fixture does via comment.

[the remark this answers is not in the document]
zmichaels11: I meant that its redundant since the name of the test fixture directly matches the comment attached to it.

dabonnie:
Build and tests pass, but codecov is still broken.

PR #73, opened 2020-01-15
Use rclcpp::NodeOptions during node construction

mm318:
In order to support composeable nodes, the configurations for the node need to come from the `rclcpp::NodeOptions` as opposed to from C++ parameters in code.

### DIFF
--- a/system_metrics_collector/src/system_metrics_collector/constants.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/constants.hpp
@@ -15,7 +15,6 @@
 #ifndef SYSTEM_METRICS_COLLECTOR__CONSTANTS_HPP_
 #define SYSTEM_METRICS_COLLECTOR__CONSTANTS_HPP_
 
-#include <chrono>
 
 namespace system_metrics_collector
 {
@@ -24,8 +23,12 @@ namespace collector_node_constants
 {
 
 constexpr const char kStatisticsTopicName[] = "system_metrics";
-constexpr const std::chrono::seconds kDefaultCollectPeriod{1};
-constexpr const std::chrono::minutes kDefaultPublishPeriod{1};
+
+constexpr const char kCollectPeriodParam[] = "measurement_period";
+constexpr const int64_t kDefaultCollectPeriodInMs = 1000;   // 1 second
+
+constexpr const char kPublishPeriodParam[] = "publish_period";
+constexpr const int64_t kDefaultPublishPeriodInMs = 60000;  // 1 minute
 
 }  // namespace collector_node_constants
 
--- a/system_metrics_collector/src/system_metrics_collector/constants.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/constants.hpp
@@ -15,6 +15,7 @@
 #ifndef SYSTEM_METRICS_COLLECTOR__CONSTANTS_HPP_
 #define SYSTEM_METRICS_COLLECTOR__CONSTANTS_HPP_
 
+#include <chrono>
 
 namespace system_metrics_collector
 {
@@ -25,10 +26,11 @@ namespace collector_node_constants
 constexpr const char kStatisticsTopicName[] = "system_metrics";
 
 constexpr const char kCollectPeriodParam[] = "measurement_period";
-constexpr const int64_t kDefaultCollectPeriodInMs = 1000;   // 1 second
+constexpr const std::chrono::milliseconds kDefaultCollectPeriod{1000};    // 1 second
 
 constexpr const char kPublishPeriodParam[] = "publish_period";
-constexpr const int64_t kDefaultPublishPeriodInMs = 60000;  // 1 minute
+constexpr const std::chrono::milliseconds kDefaultPublishPeriod{60000};   // 1 minute
+
 
 }  // namespace collector_node_constants
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_collector.cpp
@@ -31,9 +31,7 @@ int main(int argc, char ** argv)
 
   const auto cpu_node = std::make_shared<system_metrics_collector::LinuxCpuMeasurementNode>(
     "linuxCpuCollector",
-    system_metrics_collector::collector_node_constants::kDefaultCollectPeriod,
-    system_metrics_collector::collector_node_constants::kStatisticsTopicName,
-    system_metrics_collector::collector_node_constants::kDefaultPublishPeriod);
+    rclcpp::NodeOptions());
 
   rclcpp::executors::MultiThreadedExecutor ex;
   cpu_node->Start();
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_collector.cpp
@@ -30,8 +30,7 @@ int main(int argc, char ** argv)
   rclcpp::init(argc, argv);
 
   const auto cpu_node = std::make_shared<system_metrics_collector::LinuxCpuMeasurementNode>(
-    "linuxCpuCollector",
-    rclcpp::NodeOptions());
+    "linuxCpuCollector");
 
   rclcpp::executors::MultiThreadedExecutor ex;
   cpu_node->Start();
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_collector.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_collector.cpp
@@ -29,8 +29,8 @@ int main(int argc, char ** argv)
 {
   rclcpp::init(argc, argv);
 
-  const auto cpu_node = std::make_shared<system_metrics_collector::LinuxCpuMeasurementNode>(
-    "linuxCpuCollector");
+  const auto cpu_node =
+    std::make_shared<system_metrics_collector::LinuxCpuMeasurementNode>("linuxCpuCollector");
 
   rclcpp::executors::MultiThreadedExecutor ex;
   cpu_node->Start();
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
@@ -39,8 +39,9 @@ namespace system_metrics_collector
 LinuxCpuMeasurementNode::LinuxCpuMeasurementNode(
   const std::string & name,
   const rclcpp::NodeOptions & options)
-: PeriodicMeasurementNode(name, options)
-{}
+: PeriodicMeasurementNode{name, options}
+{
+}
 
 bool LinuxCpuMeasurementNode::SetupStart()
 {
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
@@ -38,10 +38,8 @@ namespace system_metrics_collector
 
 LinuxCpuMeasurementNode::LinuxCpuMeasurementNode(
   const std::string & name,
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period)
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(name, options)
 {}
 
 bool LinuxCpuMeasurementNode::SetupStart()
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
@@ -44,7 +44,7 @@ public:
    */
   LinuxCpuMeasurementNode(
     const std::string & name,
-    const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
+    const rclcpp::NodeOptions & options = rclcpp::NodeOptions{});
 
   virtual ~LinuxCpuMeasurementNode() = default;
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
@@ -37,7 +37,7 @@ public:
    * Construct a LinuxCpuMeasurementNode.
    * The following parameters may be set via the rclcpp::NodeOptions:
    * `measurement_period`: the period of this node, used to read measurements
-   * `publish_period`: the period at which metrics are published. 0 ms means don't publish
+   * `publish_period`: the period at which metrics are published
    *
    * @param name the name of this node
    * @param options the options (arguments, parameters, etc.) for this node
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
@@ -34,14 +34,17 @@ class LinuxCpuMeasurementNode : public system_metrics_collector::PeriodicMeasure
 {
 public:
   /**
-   * Construct a LinuxCpuMeasurementNode
+   * Construct a LinuxCpuMeasurementNode.
+   * The following parameters may be set via the rclcpp::NodeOptions:
+   * `measurement_period`: the period of this node, used to read measurements
+   * `publish_period`: the period at which metrics are published. 0 ms means don't publish
    *
    * @param name the name of this node
-   * @param measurement_period the period of this node, used to read measurements
-   * @param topic the topic name used for publishing
-   * @param publish_period the period at which metrics are published. 0 ms means don't publish
+   * @param options the options (arguments, parameters, etc.) for this node
    */
-  LinuxCpuMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options);
+  LinuxCpuMeasurementNode(
+    const std::string & name,
+    const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
 
   virtual ~LinuxCpuMeasurementNode() = default;
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
@@ -41,11 +41,7 @@ public:
    * @param topic the topic name used for publishing
    * @param publish_period the period at which metrics are published. 0 ms means don't publish
    */
-  LinuxCpuMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & topic,
-    const std::chrono::milliseconds publish_period);
+  LinuxCpuMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options);
 
   virtual ~LinuxCpuMeasurementNode() = default;
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.cpp
@@ -38,10 +38,8 @@ namespace system_metrics_collector
 
 LinuxMemoryMeasurementNode::LinuxMemoryMeasurementNode(
   const std::string & name,
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period)
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(name, options)
 {
 }
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.cpp
@@ -39,7 +39,7 @@ namespace system_metrics_collector
 LinuxMemoryMeasurementNode::LinuxMemoryMeasurementNode(
   const std::string & name,
   const rclcpp::NodeOptions & options)
-: PeriodicMeasurementNode(name, options)
+: PeriodicMeasurementNode{name, options}
 {
 }
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
@@ -32,13 +32,16 @@ class LinuxMemoryMeasurementNode : public system_metrics_collector::PeriodicMeas
 public:
   /**
    * Construct a LinuxMemoryMeasurementNode
+   * The following parameters may be set via the rclcpp::NodeOptions:
+   * `measurement_period`: the period of this node, used to read measurements
+   * `publish_period`: the period at which metrics are published
    *
    * @param name the name of this node
-   * @param measurement_period the period of this node, used to read measurements
-   * @param topic the topic name used for publishing
-   * @param publish_period the period at which metrics are published. 0 ms means don't publish
+   * @param options the options (arguments, parameters, etc.) for this node
    */
-  LinuxMemoryMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options);
+  LinuxMemoryMeasurementNode(
+    const std::string & name,
+    const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
 
   virtual ~LinuxMemoryMeasurementNode() = default;
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
@@ -38,11 +38,7 @@ public:
    * @param topic the topic name used for publishing
    * @param publish_period the period at which metrics are published. 0 ms means don't publish
    */
-  LinuxMemoryMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & topic,
-    const std::chrono::milliseconds publish_period);
+  LinuxMemoryMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options);
 
   virtual ~LinuxMemoryMeasurementNode() = default;
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
@@ -41,7 +41,7 @@ public:
    */
   LinuxMemoryMeasurementNode(
     const std::string & name,
-    const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
+    const rclcpp::NodeOptions & options = rclcpp::NodeOptions{});
 
   virtual ~LinuxMemoryMeasurementNode() = default;
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
@@ -42,8 +42,8 @@ namespace system_metrics_collector
 LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
   const std::string & name,
   const rclcpp::NodeOptions & options)
-: PeriodicMeasurementNode(name, options),
-  metric_name_(std::to_string(GetPid()) + kMetricName)
+: PeriodicMeasurementNode{name, options},
+  metric_name_{std::to_string(GetPid()) + kMetricName}
 {
 }
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
@@ -40,7 +40,7 @@ namespace system_metrics_collector
 {
 
 LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
-  const std::string & name, 
+  const std::string & name,
   const rclcpp::NodeOptions & options)
 : PeriodicMeasurementNode(name, options),
   metric_name_(std::to_string(GetPid()) + kMetricName)
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
@@ -40,11 +40,9 @@ namespace system_metrics_collector
 {
 
 LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
-  const std::string & name,
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period),
+  const std::string & name, 
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(name, options),
   metric_name_(std::to_string(GetPid()) + kMetricName)
 {
 }
--- a/system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
@@ -52,10 +52,8 @@ namespace system_metrics_collector
 
 LinuxProcessMemoryMeasurementNode::LinuxProcessMemoryMeasurementNode(
   const std::string & name,
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period),
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(name, options),
   pid_(std::to_string(GetPid())),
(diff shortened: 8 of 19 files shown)


mm318:
Still need to fix some minor unit test failures.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -28,34 +30,50 @@ namespace system_metrics_collector
...
-  if (measurement_period <= std::chrono::milliseconds{0}) {
+  descriptor.description = "The period in milliseconds between each measurement";
+  auto measurement_period = declare_parameter(
+    collector_node_constants::kCollectPeriodParam,
+    collector_node_constants::kDefaultCollectPeriodInMs,
+    descriptor);
+  measurement_period_ = std::chrono::milliseconds{measurement_period};
+  if (measurement_period_ <= std::chrono::milliseconds{0}) {
     throw std::invalid_argument{"measurement_period cannot be negative"};
   }
```
mm318: This is actually redundant now. `declare_parameter()` will throw before reaching this `if` statement. Thoughts on removing this check?

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -28,34 +30,50 @@ namespace system_metrics_collector
...
-  if (publish_period <= std::chrono::milliseconds{0}) {
+  descriptor.description = "The period in milliseconds between each published MetricsMessage";
+  auto publish_period = declare_parameter(
+    collector_node_constants::kPublishPeriodParam,
+    collector_node_constants::kDefaultPublishPeriodInMs,
+    descriptor);
+  publish_period_ = std::chrono::milliseconds{publish_period};
+  if (publish_period_ <= std::chrono::milliseconds{0}) {
     throw std::invalid_argument{"publish_period cannot be negative"};
   }
```
mm318: This is actually redundant now. `declare_parameter()` will throw before reaching this `if` statement. Thoughts on removing this check?

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: My understanding of the parameter descriptor is that it does not allow values to be set outside of the descriptor range. What happens if the default value also falls outside of that range. Would we get a compile or runtime error?

The only reason I see to keep the checks is to enforce sane default values, which would be found at the unit test level.

mm318: If there's no overriding value, then the descriptor will effectively check the default value. If the default value falls outside of that range, I think it will be a runtime error, but for sure we'll effectively have a broken product.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
```diff
@@ -41,11 +41,7 @@ class LinuxCpuMeasurementNode : public system_metrics_collector::PeriodicMeasure
    * @param topic the topic name used for publishing
    * @param publish_period the period at which metrics are published. 0 ms means don't publish
    */
-  LinuxCpuMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & topic,
-    const std::chrono::milliseconds publish_period);
+  LinuxCpuMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options);
```
dabonnie: The above doc isn't valid anymore. We could add expected node options and default behavior.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
thomas-moulard: I think we can remove the check as long as we have a test trying to set a bad value to make sure the param logic is correct.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
thomas-moulard: +1

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -38,10 +38,8 @@ namespace system_metrics_collector
 
 LinuxCpuMeasurementNode::LinuxCpuMeasurementNode(
   const std::string & name,
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period)
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(name, options)
```
dabonnie:
```suggestion
: PeriodicMeasurementNode{name, options}
```

system_metrics_collector/src/system_metrics_collector/main.cpp
```diff
@@ -59,29 +53,21 @@ int main(int argc, char ** argv)
...
-    kStatisticsTopicName,
-    kDefaultPublishPeriod);
+    rclcpp::NodeOptions());
 
   const auto mem_node = std::make_shared<system_metrics_collector::LinuxMemoryMeasurementNode>(
     "linuxMemoryCollector",
-    kDefaultCollectPeriod,
-    kStatisticsTopicName,
-    kDefaultPublishPeriod);
+    rclcpp::NodeOptions());
```
thomas-moulard: I think this is the default behavior, we could just pass the name instead? (ditto for all the other instances).

system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.cpp
```diff
@@ -38,10 +38,8 @@ namespace system_metrics_collector
 
 LinuxMemoryMeasurementNode::LinuxMemoryMeasurementNode(
   const std::string & name,
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period)
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(name, options)
```
dabonnie:
```suggestion
: PeriodicMeasurementNode{name, options}
```

system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
```diff
@@ -38,11 +38,7 @@ class LinuxMemoryMeasurementNode : public system_metrics_collector::PeriodicMeas
    * @param topic the topic name used for publishing
    * @param publish_period the period at which metrics are published. 0 ms means don't publish
    */
-  LinuxMemoryMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & topic,
-    const std::chrono::milliseconds publish_period);
+  LinuxMemoryMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options);
```
dabonnie: Same comment about the doc.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
```diff
@@ -41,10 +41,8 @@ namespace system_metrics_collector
 
 LinuxProcessCpuMeasurementNode::LinuxProcessCpuMeasurementNode(
   const std::string & name,
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period),
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(name, options),
```
dabonnie:
```suggestion
: PeriodicMeasurementNode{name, options},
```

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp
```diff
@@ -41,10 +41,8 @@ namespace system_metrics_collector
...
   metric_name_(std::to_string(GetPid()) + kMetricName)
```
dabonnie:
```suggestion
  metric_name_{std::to_string(GetPid()) + kMetricName}
```

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
```diff
@@ -52,10 +52,8 @@ namespace system_metrics_collector
 
 LinuxProcessMemoryMeasurementNode::LinuxProcessMemoryMeasurementNode(
   const std::string & name,
-  const std::chrono::milliseconds measurement_period,
-  const std::string & topic,
-  const std::chrono::milliseconds publish_period)
-: PeriodicMeasurementNode(name, measurement_period, topic, publish_period),
+  const rclcpp::NodeOptions & options)
+: PeriodicMeasurementNode(name, options),
```
dabonnie:
```suggestion
: PeriodicMeasurementNode{name, options},
```

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
```diff
@@ -52,10 +52,8 @@ namespace system_metrics_collector
...
   pid_(std::to_string(GetPid())),
```
dabonnie:
```suggestion
  pid_{std::to_string(GetPid())},
```

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
```diff
@@ -52,10 +52,8 @@ namespace system_metrics_collector
...
   file_to_read_(kProc + std::to_string(GetPid()) + kStatm)
```
dabonnie:
```suggestion
  file_to_read_{kProc + std::to_string(GetPid()) + kStatm}
```

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.hpp
```diff
@@ -52,11 +52,7 @@ class LinuxProcessMemoryMeasurementNode : public PeriodicMeasurementNode
    * @param topic the topic name used for publishing
    * @param publish_period the period at which metrics are published.
    */
-  LinuxProcessMemoryMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & topic,
-    const std::chrono::milliseconds publish_period);
+  LinuxProcessMemoryMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options);
```
dabonnie: Same comment concerning the doc

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -46,11 +46,7 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector,
    * strictly greater than measurement_period.
    * @throws std::invalid_argument for any invalid input
    */
-  PeriodicMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & publishing_topic,  // todo @dbbonnie think about a default topic
-    const std::chrono::milliseconds publish_period);
+  PeriodicMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options);
```
dabonnie: Doc string update

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -28,34 +30,50 @@ namespace system_metrics_collector
...
-  const std::string & publishing_topic,
-  const std::chrono::milliseconds publish_period)
-: Node(name),
-  measurement_period_(measurement_period),
-  publishing_topic_(publishing_topic),
-  measurement_timer_(nullptr),
-  publish_timer_(nullptr),
-  publish_period_(publish_period)
+  const rclcpp::NodeOptions & options)
+: Node(name, options)
```
dabonnie:
```suggestion
: Node{name, options}
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -104,7 +122,7 @@ std::string PeriodicMeasurementNode::GetStatusString() const
   std::stringstream ss;
   ss << "name=" << get_name() <<
     ", measurement_period=" << std::to_string(measurement_period_.count()) << "ms" <<
-    ", publishing_topic=" << publishing_topic_ <<
+    ", publishing_topic=" << ((publisher_ == nullptr) ? "" : publisher_->get_topic_name()) <<
```
dabonnie: The .hpp still has ```publishing_topic_```. Why tie displaying the topic to the node state?

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -108,15 +104,15 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector,
...
   /**
    * The period used to take a single measurement
    */
-  const std::chrono::milliseconds measurement_period_;
+  std::chrono::milliseconds measurement_period_;
   /**
    * The period used to publish measurement data
    */
-  const std::chrono::milliseconds publish_period_;
+  std::chrono::milliseconds publish_period_;
```
dabonnie: Is this used anymore?

system_metrics_collector/test/system_metrics_collector/test_linux_memory_measurement.cpp
```diff
@@ -89,12 +91,8 @@ constexpr const std::array<const char *, 10> kSamples = {
...
 {
 public:
-  TestLinuxMemoryMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & publishing_topic,
-    const std::chrono::milliseconds publish_period)
-  : LinuxMemoryMeasurementNode(name, measurement_period, publishing_topic, publish_period),
+  TestLinuxMemoryMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options)
+  : LinuxMemoryMeasurementNode(name, options),
```
dabonnie:
```suggestion
  : LinuxMemoryMeasurementNode{name, options},
```

system_metrics_collector/test/system_metrics_collector/test_linux_process_memory_measurement_node.cpp
```diff
@@ -35,10 +37,8 @@ class TestLinuxProcessMemoryMeasurementNode : public system_metrics_collector::
 public:
   TestLinuxProcessMemoryMeasurementNode(
     const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & topic,
-    const std::chrono::milliseconds publish_period)
-  : LinuxProcessMemoryMeasurementNode(name, measurement_period, topic, publish_period) {}
+    const rclcpp::NodeOptions & options)
+  : LinuxProcessMemoryMeasurementNode(name, options) {}
```
dabonnie:
```suggestion
  : LinuxProcessMemoryMeasurementNode{name, options} {}
```

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: Sounds good. We do have such tests.

system_metrics_collector/src/system_metrics_collector/main.cpp
mm318: I'll add a default `rclcpp::NodeOptions` parameter in the header file for each type of node.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -130,8 +142,17 @@ constexpr std::chrono::milliseconds PeriodicMeasurementTestFixure::kDontPublishD
 
 TEST_F(PeriodicMeasurementTestFixure, Sanity) {
   ASSERT_NE(test_periodic_measurer_, nullptr);
+
+  ASSERT_EQ("name=test_periodic_node, measurement_period=50ms,"
+    " publishing_topic=, publish_period=500ms, started=false,"
```
dabonnie: An empty ```publishing_topic``` isn't indicative of what was actually set. I'm not a fan of this change. If we want to tie it to publisher state then I would only display when started.

dabonnie:
> Why do we need to rename the topic? I think we can remove kTestTopic and use the real topic, no?

I agree with this, since it seems redundant to test the ROS2 parameter code.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
mm318: The value is coming from the parameters and stored here. Is your idea that the parameters values should be read at `Start()` and not stored, so that each `Stop()`->`Start()` cycle might have updated/different parameter values?

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: If displaying the topic doesn't depend on the node state, then when the user does topic name remapping, the output is technically wrong.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
dabonnie: Sorry, this is the wrong line. I meant for line 107 (```std::string publishing_topic_;```). Mainly because there is a change with the status string that does not used ```publishing_topic_```.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/main.cpp

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp

system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp

system_metrics_collector/src/system_metrics_collector/linux_process_cpu_measurement_node.cpp

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_memory_measurement_node.hpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
mm318: Okay, I removed the `publishing_topic_` member variable.

system_metrics_collector/test/system_metrics_collector/test_linux_process_memory_measurement_node.cpp
mm318: Done.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
mm318: Okay, changed this test to call `Start()` on the measurement node beforehand.

mm318:
> Why do we need to rename the topic? I think we can remove kTestTopic and use the real topic, no?

I have removed the remapping of the publishing topic to `kTestTopic`.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -28,31 +30,36 @@ namespace system_metrics_collector
...
+  positive_range.to_value =
+    std::numeric_limits<decltype(rcl_interfaces::msg::IntegerRange::to_value)>::max();
+  positive_range.step = 1;
 
-  if (publishing_topic.empty()) {
-    throw std::invalid_argument{"publishing_topic cannot be empty"};
-  }
+  rcl_interfaces::msg::ParameterDescriptor descriptor;
+  descriptor.read_only = true;
+  descriptor.integer_range.push_back(positive_range);
```
piraka9011: does using `emplace_back` make sense here?

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
piraka9011: Could this be simplified to `publisher_ ? publisher_->get_topic_name() : ""`

system_metrics_collector/test/system_metrics_collector/test_linux_memory_measurement.cpp
piraka9011: Doesn't look like the braces were applied here

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -38,19 +40,14 @@ using moving_average_statistics::StatisticData;
...
 {
 public:
-  MockLinuxProcessCpuMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & publishing_topic,
-    const std::chrono::milliseconds publish_period)
-  : LinuxProcessCpuMeasurementNode(name, measurement_period, publishing_topic, publish_period) {}
+  MockLinuxProcessCpuMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options)
+  : LinuxProcessCpuMeasurementNode(name, options) {}
```
piraka9011: Ditto above, brace init.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -171,34 +182,83 @@ TEST_F(PeriodicMeasurementTestFixure, TestStartAndStop) {
 }
 
 TEST_F(PeriodicMeasurementTestFixure, TestConstructorInputValidation) {
+  std::vector<rclcpp::Parameter> parameter_overrides;
+  std::vector<std::string> arguments;
+
+  // bad measurement period
+  parameter_overrides.push_back(rclcpp::Parameter(
```
piraka9011: Is it better to have these tests separated?

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -42,12 +44,9 @@ constexpr const char kTestMetricname[] = "test_metric_name";
...
 {
 public:
-  TestPeriodicMeasurementNode(
-    const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & publishing_topic,
-    const std::chrono::milliseconds publish_period)
-  : PeriodicMeasurementNode(name, measurement_period, publishing_topic, publish_period) {}
+  TestPeriodicMeasurementNode(const std::string & name, const rclcpp::NodeOptions & options)
+  : PeriodicMeasurementNode(name, options) {}
```
piraka9011: Ditto above, brace init

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
```diff
@@ -68,7 +75,8 @@ bool PeriodicMeasurementNode::SetupStart()
     measurement_period_, [this]() {this->PerformPeriodicMeasurement();});
 
   if (publisher_ == nullptr) {
-    publisher_ = create_publisher<MetricsMessage>(publishing_topic_, 10 /*history_depth*/);
+    publisher_ = create_publisher<MetricsMessage>(collector_node_constants::kStatisticsTopicName,
```
dabonnie: Oh I see what you're doing now. Why aren't we keeping the member? This PR is actually removing functionality where we could pull the publishing topic out of NodeOptions. The work required seems minimal and I don't see why we can't do it, as a read only parameter, here.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
zmichaels11: I think the code would be:
```c++
descriptor.integer_range.emplace_back(
  /* from_value */ 1,
  /* to value */ std::numeric_limits<decltype(rcl_interfaces::msg::IntegerRange::to_value)>::max(),
  /* step value */ 1);
```
an alternative approach would be to use the `rvalue push_back` by doing:
```c++
descriptor.integer_range.push_back(std::move(positive_range));
```

IMO: readability is preferred over saving the cost of initializing+moving a struct

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: > This PR is actually removing functionality where we could pull the publishing topic out of NodeOptions.

I was going to do that, too, but this is what @thomas-moulard suggested under on the same rationale as the node name.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: I'll try, but if I recall correctly I typically find that these ROS `msg` objects typically can't be constructed that way.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: Done.

system_metrics_collector/test/system_metrics_collector/test_linux_memory_measurement.cpp
mm318: Now it's really done :)

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
mm318: Done.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_cpu_collector.cpp
```diff
@@ -30,10 +30,7 @@ int main(int argc, char ** argv)
   rclcpp::init(argc, argv);
 
   const auto cpu_node = std::make_shared<system_metrics_collector::LinuxCpuMeasurementNode>(
```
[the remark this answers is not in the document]
mm318: Done.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
```diff
@@ -34,18 +34,17 @@ class LinuxCpuMeasurementNode : public system_metrics_collector::PeriodicMeasure
...
-   * @param topic the topic name used for publishing
-   * @param publish_period the period at which metrics are published. 0 ms means don't publish
+   * @param options the options (arguments, parameters, etc.) for this node
    */
   LinuxCpuMeasurementNode(
     const std::string & name,
-    const std::chrono::milliseconds measurement_period,
-    const std::string & topic,
-    const std::chrono::milliseconds publish_period);
+    const rclcpp::NodeOptions & options = rclcpp::NodeOptions());
```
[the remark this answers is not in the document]
mm318: Done.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
mm318: Done.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
dabonnie: I took @thomas-moulard 's comment to mean we don't need to set and test a custom topic name in test. I'm going to follow up with him as we've discussed allowing a user to configure the topic name.

system_metrics_collector/src/system_metrics_collector/constants.hpp
```diff
@@ -25,10 +26,11 @@ namespace collector_node_constants
 constexpr const char kStatisticsTopicName[] = "system_metrics";
 
 constexpr const char kCollectPeriodParam[] = "measurement_period";
-constexpr const int64_t kDefaultCollectPeriodInMs = 1000;   // 1 second
+constexpr const std::chrono::milliseconds kDefaultCollectPeriod{1000};    // 1 second
```
zmichaels11: Why is this still in milliseconds when the delay is now 1 second?

mm318: I need to get the `count()` in milliseconds at some point.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.cpp
mm318: This was the discussion in the team chatroom. Topic name will be configured via remapping (`--ros-args --remap ...`).

dabonnie: > --ros-args --remap

Got it, found this explanation: https://index.ros.org/doc/ros2/Tutorials/Node-arguments/#example

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
```diff
@@ -34,18 +34,17 @@ class LinuxCpuMeasurementNode : public system_metrics_collector::PeriodicMeasure
 {
 public:
   /**
-   * Construct a LinuxCpuMeasurementNode
+   * Construct a LinuxCpuMeasurementNode.
+   * The following parameters may be set via the rclcpp::NodeOptions:
+   * `measurement_period`: the period of this node, used to read measurements
+   * `publish_period`: the period at which metrics are published. 0 ms means don't publish
```
dabonnie:
```suggestion
   * `publish_period`: the period at which metrics are published.
```

Nit: given the input argument checking, 0 ms isn't valid.

dabonnie: Same comment per the 0ms docstring.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -170,35 +180,69 @@ TEST_F(PeriodicMeasurementTestFixure, TestStartAndStop) {
...
+    std::chrono::milliseconds{-1}.count());
+  options.append_parameter_override(
+    system_metrics_collector::collector_node_constants::kPublishPeriodParam,
+    kDontPublishDuringTest.count());
+
+  ASSERT_THROW(TestPeriodicMeasurementNode("throw", options),
+    rclcpp::exceptions::InvalidParameterValueException);
+}
+
+TEST_F(PeriodicMeasurementTestFixure, TestConstructorMeasurementPeriodValidation2) {
```
dabonnie: How is this not identical to TestConstructorMeasurementPeriodValidation1?

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.hpp
mm318: Done.

mm318: This is the same comment as https://github.com/ros-tooling/system_metrics_collector/pull/73#discussion_r367700566.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
mm318: I have removed `TestConstructorMeasurementPeriodValidation1`.

dabonnie:
Build successful but codecov still broken. FYI @ros2/aws-oncall

PR #75, opened 2020-01-17
Added some doc to system_metric_collector

ahcorde:
Added some doc to system_metric_collector

### DIFF
--- a/system_metrics_collector/README.md
+++ b/system_metrics_collector/README.md
@@ -1,0 +1,79 @@
+# System metric collector
+
+## Parameters
+
+There are two parameters defined:
+
+  - **measurement_period**:
+```sh
+Parameter name: measurement_period
+Type: integer
+  Description: The period in milliseconds between each measurement
+  Constraints:
+    Read only: true
+    Min value: 1
+    Max value: 9223372036854775807
+    Step: 1
+```
+  - **publish_period**:
+```sh
+Parameter name: publish_period
+  Type: integer
+  Description: The period in milliseconds between each published MetricsMessage
+  Constraints:
+    Read only: true
+    Min value: 1
+    Max value: 9223372036854775807
+    Step: 1
+```
+
+## Usage:
+
+You can run the following command to generate CPU and Memory statistics. for both, all the system and the node running:
+
+```
+ros2 run system_metrics_collector main
+```
+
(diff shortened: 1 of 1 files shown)


system_metrics_collector/README.md
```diff
@@ -0,0 +1,79 @@
+# System metric collector
+
+## Parameters
+
+There are two parameters defined:
+
+  - **measurement_period**:
+```sh
```
piraka9011: consider using `yaml` better readability

system_metrics_collector/README.md
```diff
@@ -0,0 +1,79 @@
...
+  Constraints:
+    Read only: true
+    Min value: 1
+    Max value: 9223372036854775807
+    Step: 1
+```
+
+## Usage:
+
+You can run the following command to generate CPU and Memory statistics. for both, all the system and the node running:
```
piraka9011: Remove the second sentence, you only need the first

system_metrics_collector/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+```
+ros2 run system_metrics_collector main
+```
+
+Change `publish_period` or `measurement_period` using `--ros-args`:
```
piraka9011: `publish_period` or `measurement_period`  can be modified using...

system_metrics_collector/README.md
```diff
@@ -0,0 +1,79 @@
...
+    Max value: 9223372036854775807
+    Step: 1
+```
+
+## Usage:
+
+You can run the following command to generate CPU and Memory statistics. for both, all the system and the node running:
+
+```
+ros2 run system_metrics_collector main
```
dabonnie: This usage isn't quite correct. At some point in the next 2 weeks, we will be moving this main to an example directory. The preferred way to launch the system CPU and system memory collector will be to use their entry points (standalone mains) or the launch file. Per process CPU and memory will use composition.

thomas-moulard: @dabonnie Is this the correct usage now? If yes, I think we can merge as it is and the refactoring PR should update the doc accordingly at this time?

dabonnie: @thomas-moulard sounds good, why I didn't submit as a non-blocking review ;-)

dabonnie:
Merging as build passes but codecov is broken.

PR #76, opened 2020-01-20
Add unit test for composeability of nodes

mm318:
This adds tests for whether or not `libsystem_metrics_collector.so` contains working composeable nodes. It is based on one of the [composition demos](https://github.com/ros2/demos/blob/master/composition/src/dlopen_composition.cpp).

### DIFF
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -28,7 +28,6 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
-find_package(class_loader REQUIRED)
 find_package(metrics_statistics_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_components REQUIRED)
@@ -79,6 +78,7 @@ ament_target_dependencies(linux_memory_collector)
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   find_package(ament_cmake_gtest REQUIRED)
+  find_package(class_loader REQUIRED)
   find_package(lifecycle_msgs REQUIRED)
 
   ament_lint_auto_find_test_dependencies()
--- a/system_metrics_collector/CMakeLists.txt
+++ b/system_metrics_collector/CMakeLists.txt
@@ -28,6 +28,7 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
+find_package(class_loader REQUIRED)
 find_package(metrics_statistics_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rclcpp_components REQUIRED)
@@ -121,6 +122,10 @@ if(BUILD_TESTING)
     test/system_metrics_collector/test_periodic_measurement_node.cpp)
   target_link_libraries(test_periodic_measurement_node system_metrics_collector)
   ament_target_dependencies(test_periodic_measurement_node lifecycle_msgs metrics_statistics_msgs rclcpp)
+
+  ament_add_gtest(test_composition
+          test/system_metrics_collector/test_composition.cpp)
+  ament_target_dependencies(test_composition class_loader rclcpp rclcpp_components)
 
   ament_add_gtest(test_utilities
           test/system_metrics_collector/test_utilities.cpp)
--- a/system_metrics_collector/test/system_metrics_collector/test_composition.cpp
+++ b/system_metrics_collector/test/system_metrics_collector/test_composition.cpp
@@ -1,0 +1,76 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <string>
+#include <vector>
+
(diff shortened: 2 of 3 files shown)


system_metrics_collector/test/system_metrics_collector/test_composition.cpp
```diff
@@ -0,0 +1,76 @@
...
+constexpr const char kSystemMetricsCollectorLibName[] = "libsystem_metrics_collector.so";
+constexpr const std::array<const char *, 2> kExpectedClassNames = {
+  "system_metrics_collector::LinuxProcessCpuMeasurementNode",
+  "system_metrics_collector::LinuxProcessMemoryMeasurementNode"
+};
+}
+
+bool IsExpectedClassName(const std::string & class_name)
+{
+  for (auto expected_class_name : kExpectedClassNames) {
```
piraka9011: Why not use `std::find` instead of using a loop?
```c++
auto result = std::find(kExpectedClassNames.begin(), kExpectedClassNames.end(), class_name);
return result != kExpectedClassNames.end();
```

mm318: The elements in `kExpectedClassNames` will actually be substrings of `class_name`.

mm318: But sure, I can replace the `for` loop with a `std::find_if()`.

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
```diff
@@ -0,0 +1,76 @@
...
+  return false;
+}
+
+TEST(TestComposeableNodes, DlopenTest)
+{
+  rclcpp::init(0, nullptr);
+  rclcpp::executors::SingleThreadedExecutor exec;
+  rclcpp::NodeOptions options;
+
+  auto loader = std::make_unique<class_loader::ClassLoader>(kSystemMetricsCollectorLibName);
```
zmichaels11: loader can be `const`

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  std::promise<bool> empty_promise;
+  std::shared_future<bool> dummy_future = empty_promise.get_future();
+  exec.spin_until_future_complete(dummy_future, test_constants::kTestDuration);
+
+  for (auto & wrapper : node_wrappers) {
+    exec.remove_node(wrapper.get_node_base_interface());
+  }
+  node_wrappers.clear();
```
zmichaels11: ~~I don't think `vector::clear` actually deconstructs the held items.~~
I think it would be better to open a scope from where you define `node_wrappers` (L56) to L72 in place of calling `clear` so `node_wrappers` and its held elements do get deconstructed when it goes out of scope.

Edit:
`std::vector::clear` [does deconstruct all held elements](https://stackoverflow.com/questions/22709344/does-stdvectortclear-call-destructor-of-the-content). I still think it looks cleaner by letting it go out of scope.

mm318: Then I'll remove the line completely and let `node_wrappers` go out of the scope of the test body.

system_metrics_collector/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ endif()
 
 # find dependencies
 find_package(ament_cmake REQUIRED)
+find_package(class_loader REQUIRED)
```
dabonnie: Is this needed as a build / executable dependency or just test? 

Doesn't this requirement also need to be added to package.xml?

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+TEST(TestComposeableNodes, DlopenTest)
+{
+  rclcpp::init(0, nullptr);
+  rclcpp::executors::SingleThreadedExecutor exec;
+  rclcpp::NodeOptions options;
+
+  const auto loader = std::make_unique<class_loader::ClassLoader>(kSystemMetricsCollectorLibName);
+  auto class_names = loader->getAvailableClasses<rclcpp_components::NodeFactory>();
```
dabonnie: const?

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
```diff
@@ -0,0 +1,72 @@
...
+  ASSERT_EQ(kExpectedClassNames.size(), class_names.size());
+
+  std::vector<rclcpp_components::NodeInstanceWrapper> node_wrappers;
+  for (const auto & class_name : class_names) {
+    ASSERT_TRUE(IsExpectedClassName(class_name));
+    auto node_factory = loader->createInstance<rclcpp_components::NodeFactory>(class_name);
```
dabonnie: const?

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto wrapper = node_factory->create_node_instance(options);
```
dabonnie: const?

system_metrics_collector/CMakeLists.txt
mm318: Hmm, it's strange that our CI managed to catch missing requirements in `package.xml` for `rclcpp_components`, but not for `class_loader`.

I'll add `class_loader` as a `test_depend` in our `package.xml`.

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
mm318: Done.

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
mm318: Done.

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
mm318: Done.

dabonnie:
Merged as build + test passed, but codecov is still broken.

PR #78, opened 2020-01-22
Fix codecov pipeline

piraka9011:
Testing [ros-tooling/action-ros-ci#63](https://github.com/ros-tooling/action-ros-ci/pulls/63) and [ros-tooling/setup-ros#46](https://github.com/ros-tooling/setup-ros/pull/46).


thomas-moulard:
Thanks for looking into this!

PR #80, opened 2020-01-22
Add configurable launch parameters

mm318:
The launch parameters may now be specified in the launch file, and consequently on the command line as well. Node names, measurement period, publish period, and publish topic are things that are configurable.

Example usage and output:
```
# ros2 launch system_metrics_collector system_cpu_and_memory.launch.py 
[INFO] [launch]: All log files can be found below /root/.ros/log/2020-01-22-22-04-56-835420-u0fc025e8e7ed53.ant.amazon.com-10207
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [linux_cpu_collector-1]: process started with pid [10218]
[INFO] [linux_memory_collector-2]: process started with pid [10219]
[linux_cpu_collector-1] [ERROR] [1579730697.973499051] [ComputeCpuActivePercentage]: a measurement was empty, unable to compute cpu percentage
...
```


system_metrics_collector/launch/system_cpu_and_memory.launch.py
```diff
@@ -12,24 +12,68 @@
...
+"""Launch system CPU and system memory metrics collector nodes."""
 def generate_launch_description():
+    # Default to included config file
+    config_filepath = os.path.join(
+        get_package_share_directory('system_metrics_collector'),
+        'config',
+        'system_metrics_collector_config.yaml')
+    with open(config_filepath, 'r') as f:
+        config_file_contents = f.read()
+    config_yaml_data = yaml.safe_load(config_file_contents)
```
dabonnie: Can't this throw if the config contains invalid yaml?

mm318: So what do you want the behavior of an invalid YAML to be?

mm318: The reading of a YAML file has been removed altogether.

mm318:
> Launch files should not contain any logic, let's chat about this pr

Finished chatting. Please re-review, @dabonnie and @thomas-moulard.

system_metrics_collector/launch/system_cpu_and_memory.launch.py
```diff
@@ -12,24 +12,75 @@
...
+MEMORY_NODE_NAME = 'linux_memory_collector_node_name'
+MEASUREMENT_PERIOD = 'measurement_period'
+PUBLISH_PERIOD = 'publish_period'
+PUBLISH_TOPIC = 'publish_topic'
+
+# Default argument values
+default_cpu_node_name = 'linux_cpu_collector'
+default_memory_node_name = 'linux_memory_collector'
+default_measurement_period = '1000'
+default_publish_period = '60000'
```
thomas-moulard: variables containing a scientific value should include the unit `default_measurement_period` -> `default_measurement_period_in_ms`

emersonknapp:
@thomas if i understand correctly, you are asking about this https://index.ros.org/doc/ros2/Tutorials/Node-arguments/#name-remapping

thomas-moulard:
@emersonknapp wrong username :p Yes exactly, but does it work with `roslaunch2`? I know we can rename nodes if we invoke them directly on the command line, but not sure if there is a "default" roslaunch2 arg for that.

mm318:
I think you mean something like overriding a builtin `__node` parameter, but I'm not sure how that will behave for a launch file that's launching multiple nodes (e.g.: which node's name will it change?)

emersonknapp:
Hm, yes, I'm not sure. I'm pretty sure if you know the name, you can remap _any_ ROS2 name from the launch command line, but I'm not sure how

thomas-moulard:
^ what I'm talking about - if we could educate ourselves on this now, that would solve us a lot of work in the future.

mm318:
After doing some testing, this is my understanding.

`__node` is a parameter that you may override when doing `ros2 run` to change the name of the node. In the context of a launch file, when you do `launch_ros.actions.Node(node_name=<something>)`, it is essentially doing the same thing under the hood (i.e.: adding `-r __node:=<something>` to the run parameters).

However, if you try to override `__node` by doing `ros2 launch <package> <launch_file> __node:=<something>`, it is actually the `launch_ros_[xxxx]` node (so apparently `ros2 launch` spawns an extra node) that gets renamed to `<something>` (and you'll get a warning that this is a deprecated syntax, suggesting you to do `--ros-args --remap __node:=<something> instead`, but doing that has no effect whatsoever given how our launch file is written).

mm318:
In short, I believe what we're doing in this pull request is the right approach.

PR #85, opened 2020-01-28
Add unit information to MetricsMessage

mm318:
Add a field containing unit information to `MetricsMessage`. This makes `MetricsMessage` more compatible with Cloudwatch Metrics' `MetricDatum`, but it's meaningful information to have anyways in its own right.

### DIFF
--- a/metrics_statistics_msgs/msg/MetricsMessage.msg
+++ b/metrics_statistics_msgs/msg/MetricsMessage.msg
@@ -14,8 +14,11 @@
 # Name metric measurement source, e.g., node, topic, or process name
 string measurement_source_name
 
-# Name of the metric being measured, e.g. cpu_percentage, free_memory_mb, message_age, etc
+# Name of the metric being measured, e.g. cpu_percentage, free_memory_mb, message_age, etc.
 string metrics_source
+
+# Unit of measure of the metric, e.g. percent, mb, seconds, etc.
+string unit
 
 # Measurement window start time
 builtin_interfaces/Time window_start
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
@@ -18,9 +18,10 @@
 #include <sstream>
 #include <string>
 
-#include "../../src/system_metrics_collector/linux_cpu_measurement_node.hpp"
-#include "../../src/system_metrics_collector/periodic_measurement_node.hpp"
-#include "../../src/system_metrics_collector/utilities.hpp"
+#include "constants.hpp"
+#include "linux_cpu_measurement_node.hpp"
+#include "periodic_measurement_node.hpp"
+#include "utilities.hpp"
 
 #include "rclcpp/rclcpp.hpp"
 #include "rcutils/logging_macros.h"
@@ -87,7 +88,7 @@ std::string LinuxCpuMeasurementNode::GetMetricName() const
 
 const std::string & LinuxCpuMeasurementNode::GetMetricUnit() const
 {
-  static const std::string unit_name{"percent"};
+  static const std::string unit_name{collector_node_constants::kPercentUnitName};
   return unit_name;
 }
 
--- a/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
+++ b/system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
@@ -85,4 +85,10 @@ std::string LinuxCpuMeasurementNode::GetMetricName() const
   return kMeasurementType;
 }
 
+const std::string & LinuxCpuMeasurementNode::GetMetricUnit() const
+{
+  static const std::string unit_name{"percent"};
+  return unit_name;
+}
+
 }  // namespace system_metrics_collector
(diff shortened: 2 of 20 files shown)


system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp
```diff
@@ -85,4 +85,10 @@ std::string LinuxCpuMeasurementNode::GetMetricName() const
   return kMeasurementType;
 }
 
+const std::string & LinuxCpuMeasurementNode::GetMetricUnit() const
+{
+  static const std::string unit_name{"percent"};
```
dabonnie: Since the same string is used elsewhere it could be a constant.

metrics_statistics_msgs/msg/MetricsMessage.msg
```diff
@@ -14,9 +14,12 @@
 # Name metric measurement source, e.g., node, topic, or process name
 string measurement_source_name
 
-# Name of the metric being measured, e.g. cpu_percentage, free_memory_mb, message_age, etc
+# Name of the metric being measured, e.g. cpu_percentage, free_memory_mb, message_age, etc.
 string metrics_source
 
+# Unit of measure of the metric, e.g. percent, mb, seconds, etc.
```
timrobotson: Will this be something other packages will specify/use? My first worry is that these units will be represented in different ways when it's a free form string (mb/s vs MB/sec and mb vs MB vs MiB) and plurality. If this is only used for this package it should be ok maybe with a spot in the README or documentation that describes the different units of measure used. An enum could be even better but that may be too restrictive/overkill for this.

timrobotson: Ah I see in the other PR you're using https://github.com/aws/aws-sdk-cpp/blob/master/aws-cpp-sdk-monitoring/include/aws/monitoring/model/StandardUnit.h for units. Perhaps that should be noted somewhere or used somehow in this package.

mm318: I envision most of the consumers of this type of message will be designed to take in a free-form unit name.

However, Cloudwatch Metrics is not one of them, but it is within our control. `"percent"` does happen to be a compatible unit name: https://github.com/aws-robotics/cloudwatch-common/pull/48/files#diff-4a04cf878bd9a5f882f2d615a0017b03R33.

dabonnie: Agree that we should not depend directly on CloudWatch units.

system_metrics_collector/src/system_metrics_collector/linux_cpu_measurement_node.cpp

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
```diff
@@ -92,7 +94,7 @@ std::string LinuxProcessMemoryMeasurementNode::GetMetricName() const
 
 const std::string & LinuxProcessMemoryMeasurementNode::GetMetricUnit() const
 {
-  static const std::string unit_name{"percent"};
+  static const std::string unit_name{collector_node_constants::kPercentUnitName};
   return unit_name;
```
zmichaels11: I think its going to be better to return an implicitly converted `std::string` copy of `kPercentUnitName` by value. This creates storage for `unit_name` that has a lifecycle tied to the instance of `LinuxProcessMemoryMeasurementNode`. Anything that calls `LinuxProcessMemoryMeasurementNode::GetMetricUnit()` must ensure that the returned value does not outlive the instance of `LinuxProcessMeasurementNode` otherwise there is a memory error.

The compiler will probably optimiz

system_metrics_collector/test/system_metrics_collector/test_linux_memory_measurement.cpp
```diff
@@ -46,6 +46,7 @@ namespace
 {
 constexpr const char kTestNodeName[] = "test_measure_linux_memory";
 constexpr const char kTestMetricName[] = "system_memory_percent_used";
+constexpr const char kTestMetricUnit[] = "percent";
```
dabonnie: Nit: this is used more than once throughout the tests, why not define in test constants?

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
mm318: Are you saying we should be lifecycling/churning `std::string` objects for the same constant value (`"percent"`)?

> This creates storage for `unit_name` that has a lifecycle tied to the instance of `LinuxProcessMemoryMeasurementNode`.

This is not true as `unit_name` is declared `static`.

system_metrics_collector/test/system_metrics_collector/test_linux_memory_measurement.cpp
mm318: How about just using what is in `constants.hpp` instead of adding another copy to `test_constant.hpp`?

system_metrics_collector/src/system_metrics_collector/linux_process_memory_measurement_node.cpp
zmichaels11: I got mixed up here, this is fine.

PR #88, opened 2020-01-31
Add e2e test workflow

dabonnie:


system_metrics_collector/package.xml
```diff
@@ -16,14 +16,19 @@
...
-  <!--Required for example launch file-->
-  <exec_depend>demo_nodes_cpp</exec_depend>
   <exec_depend>metrics_statistics_msgs</exec_depend>
   <exec_depend>rclcpp</exec_depend>
   <exec_depend>rclcpp_components</exec_depend>
   <exec_depend>rclcpp_lifecycle</exec_depend>
   <exec_depend>rcpputils</exec_depend>
   <exec_depend>rcutils</exec_depend>
+  <!--Required for e2e test file-->
+  <exec_depend>rclpy</exec_depend>
```
emersonknapp: Should it be test_depend for all these?

test/requirements.txt
```diff
@@ -0,0 +1 @@
+retry
```
emersonknapp: can this be handled via rosdep via `test_depend`?

test/run_e2e_test.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# This script sources the built ROS2 workspace, installs the necessary Python dependencies,
+# and executes the python end to end tests for the system_metrics_collector package.
+
+ROS_WS_DIR=ros_ws/install/
```
emersonknapp: Is there a way to invoke this through `colcon test` instead of an external mechanism?

test/run_e2e_test.py
```diff
@@ -263,4 +268,5 @@ def main(args=None) -> int:
 if __name__ == '__main__':
     setup_logger()
     test_output = main()
+    logging.debug("exiting with test_output=%s", test_output)
```
piraka9011: Why not use `.format()`?

system_metrics_collector/package.xml
dabonnie: Isn't test depend only for [unit tests](http://wiki.ros.org/catkin/package.xml#Dependencies)?

test/run_e2e_test.sh
dabonnie: I'm not sure, but also unclear why we would need to run via ```colcon test```.

test/run_e2e_test.py
dabonnie: Simply using the logging API: https://docs.python.org/3/library/logging.html#logging.Logger.debug

test/run_e2e_test.sh
emersonknapp: General reasoning is that that's how tests are executed in a ROS2 package

system_metrics_collector/package.xml
emersonknapp: I'm not sure we should be making a technical distinction between unit tests and integration tests - only a semantic one

emersonknapp: e.g. see https://github.com/ros2/launch/tree/master/launch_testing for one of the tools that is used to test launchfiles in packages. In ROS1, we used http://wiki.ros.org/rostest to launch multi-node test environments, but I think that's easier to just do now via regular channels

test/run_e2e_test.sh
```diff
@@ -0,0 +1,15 @@
...
+TEST_DIR=ros_ws/src/system_metrics_collector/test
+
+# install the e2e test requirements
+pip3 install -r $TEST_DIR/requirements.txt
+
+# source the ROS2 installation
+source $ROS_WS_DIR/setup.bash && source $ROS_WS_DIR/local_setup.bash
+
+# run the e2e test
+python3 $TEST_DIR/run_e2e_test.py
```
thomas-moulard: nit: newline.

test/run_e2e_test.sh
thomas-moulard: Emerson, we want to avoid that because the goal is to share this shell script between canaries (i.e. getting the binary through apt), and local tests.

thomas-moulard: `ros2 pkg prefix system_metrics_collector` ?

test/run_e2e_test.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# This script sources the built ROS2 workspace, installs the necessary Python dependencies,
+# and executes the python end to end tests for the system_metrics_collector package.
+
```
thomas-moulard: Please some more doc around the fact that the objective is for this script to be able to run independently tests on APT packages (canary), and locally. I.e. Emerson was confused, it'd be good to guide the reader.

test/requirements.txt
thomas-moulard: +1 - I would declare everything as `test_depend`, and let the shell run `rosdep` to install the missing packages?

test/run_e2e_test.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
thomas-moulard: Please run shellcheck on this script and fix the lint errors, quoting seems to be unsufficient.

test/run_e2e_test.sh
thomas-moulard: You probably want `set -euf -o pipefail` here.

test/run_e2e_test.sh
dabonnie: If we want to use ```colcon test``` to execute the e2e test (intially specified as quality control for this repo via GH Actions), then I think this should be implemented as a separate PR. Seems like it could be a good first issue for an interested contributor, otherwise we're needlessly blocking.

emersonknapp:
To collect discussion:

We can leave the "e2e test as a colcon test" part out of this discussion, agreed, out of scope.

2 Remaining concerns:

1. Adding dependencies required for the canary test as `exec_depends` includes unecessary packages for all users, who may not have been installing the CLI or launch, for example.

It seems right to make them `test_depends`. To make sure we have the packages in the test environment, we have the following resources in the Github Action environment
* `apt`-installing the package installed all runtime dependencies
* We check out the package sources. We can then run `rosdep install` against the source directory to get the test dependencies

2. Is this test running a script from the master branch of this repo against the apt binary? If so, once development is done on this repo that has not been released, or if there are multiple releases out on apt, e.g, for different rosdistros, there is a mismatch that probably triggers canary failure.

I think this second one is solvable by checking the release version of the apt package that we're installing, and explicitly checking out the source tag corresponding to the release that created that deb.

We could probably duck under this second issue for a while but it seems likely to bite us unexpectedly at some point.

emersonknapp:
To extend on above - perhaps we try out that workflow for this package and if it seems to be working well we can factor it out into a more generic `apt-canary` github action that all our packages could utilize. Most of the tools seem like they'd be common, e.g. `launch` and `ros2cli`, have the specific package provide a config or script to tailor to the specific test

test/run_e2e_test.sh
dabonnie: There are errors when sourcing setup.bash, so this exits prematurely.

test/run_e2e_test.sh
dabonnie: Done

dabonnie:
Latest change refactors to use python3-retrying, as it's available via apt.

Note: python3-retrying is current not available via rosdep. Will update this PR when a PR has been opened against rosdistro.

dabonnie:
Filed https://github.com/ros-tooling/aws-roadmap/issues/182 to capture hallway discussion and comments here.

dabonnie:
Opened PR to resolve dependency in rosdistro: https://github.com/ros/rosdistro/pull/23730

test/run_e2e_test.sh
thomas-moulard: That sounds problematic - what types of errors?

emersonknapp: There are a few unset variables in colcon's `setup.bash` that allow for enabling some types of debug output. You can drop the `-u` and it should be fine

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
```diff
@@ -35,10 +35,11 @@ constexpr const std::array<const char *, 2> kExpectedClassNames = {
 
 bool IsExpectedClassName(const std::string & class_name)
 {
-  auto result = std::find_if(kExpectedClassNames.cbegin(), kExpectedClassNames.cend(),
-      [&class_name](const char * expected_class_name) {
-        return class_name.find(expected_class_name) != std::string::npos;
-      });
+  auto result = std::find_if(
```
zmichaels11:
```c++
const auto result = std::find_if(
```
?

test/run_e2e_test.sh
thomas-moulard: I agree with Emeron's recommendation. Let's make sure we explicitly check that the variables we need are set then.

system_metrics_collector/test/system_metrics_collector/test_composition.cpp
dabonnie: Done

dabonnie:
@thomas-moulard 

>I agree with Emeron's recommendation. Let's make sure we explicitly check that the variables we need are set then.

This was done via the test dependencies added in [package.xml](https://github.com/ros-tooling/system_metrics_collector/blob/ca2701c0cdcc89edc84fa88c26c1009be19f633f/system_metrics_collector/package.xml). The docstring was stale, so updated accordingly.

thomas-moulard:
@dabonnie I mean checking that the shell variables are non empty.

E.g. `[[ -z ${FOO} ]] && (echo "foo is not set"; exit 1)`

PR #93, opened 2020-02-24
Add test for collector state

dabonnie:


system_metrics_collector/test/system_metrics_collector/test_collector.cpp
```diff
@@ -84,6 +84,12 @@ TEST_F(CollectorTestFixure, Sanity) {
   ASSERT_NE(test_collector_, nullptr);
 }
 
+TEST_F(CollectorTestFixure, TestAlreadyStarted) {
+  ASSERT_NE(test_collector_, nullptr);
+  ASSERT_TRUE(test_collector_->Start());
+  ASSERT_FALSE(test_collector_->Start());
```
zmichaels11: I suggest adding a message to these assert statements so the test explains why it failed if it fails.

PR #98, opened 2020-02-27
Improve e2e message publication test

dabonnie:

### DIFF
--- a/test/run_e2e_test.py
+++ b/test/run_e2e_test.py
@@ -49,6 +49,8 @@ PUBLICATION_TEST_TIMEOUT_SECONDS = 180
 DEFAULT_MAX_ATTEMPTS = 10
 DEFAULT_WAIT_EXPONENTIAL_MULTIPLIER = 1000
 DEFAULT_MAX_EXPONENTIAL_WAIT_MILLISECONDS = 60000
+# this is longer than PUBLICATION_TEST_TIMEOUT_SECONDS in order to let the spin
+# timout complete
 DEFAULT_FIXED_WAIT_MILLISECONDS = (PUBLICATION_TEST_TIMEOUT_SECONDS + TIMEOUT_SECONDS) * 1000
 
 
@@ -69,6 +71,7 @@ class StatisticsListener(Node):
                                             EXPECTED_TOPIC,
                                             self.listener_callback,
                                             QOS_DEPTH)
+        # used to stop spinning upon success
         self.future = future
         # key is node name, value is expected number of messages to receive
         self.expected_lifecycle_nodes_dict = {}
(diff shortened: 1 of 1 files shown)


test/run_e2e_test.py
```diff
@@ -57,37 +61,58 @@ class SystemMetricsEnd2EndTestException(Exception):
...
             logging.debug('received message from %s', node_name)
-            self.expected_lifecycle_nodes.remove(node_name)
-
-        if not self.expected_lifecycle_nodes:
+            removed = False
+            with self.lock:
+                self.expected_lifecycle_nodes_dict[node_name] = (
+                        self.expected_lifecycle_nodes_dict[node_name] - 1)
+                if self.expected_lifecycle_nodes_dict[node_name] == 0:
+                    del self.expected_lifecycle_nodes_dict[node_name]
```
piraka9011: Is this necessary? Couldn't you just check for all keys in the dict that their value is 0?

test/run_e2e_test.py
```diff
@@ -57,37 +61,58 @@ class SystemMetricsEnd2EndTestException(Exception):
...
 
-        if node_name in self.expected_lifecycle_nodes:
+        if node_name in self.expected_lifecycle_nodes_dict:
             logging.debug('received message from %s', node_name)
-            self.expected_lifecycle_nodes.remove(node_name)
-
-        if not self.expected_lifecycle_nodes:
+            removed = False
+            with self.lock:
+                self.expected_lifecycle_nodes_dict[node_name] = (
```
piraka9011: Should this be a tuple?

test/run_e2e_test.py
```diff
@@ -193,28 +226,32 @@ def check_for_expected_topic(expected_topic: str) -> None:
         raise SystemMetricsEnd2EndTestException('Unable to find expected topic: ' + str(output))
 
 
-@retry(stop_max_attempt_number=DEFAULT_MAX_ATTEMPTS,
-       wait_exponential_multiplier=DEFAULT_WAIT_EXPONENTIAL_MULTIPLIER,
-       wait_exponential_max=DEFAULT_MAX_EXPONENTIAL_WAIT_MILLISECONDS)
+@retry(stop_max_attempt_number=DEFAULT_MAX_ATTEMPTS, wait_fixed=DEFAULT_FIXED_WAIT_MILLISECONDS)
```
piraka9011: Is the wait no longer needed b/c you're already checking the expected number of msgs to receive?

test/run_e2e_test.py
dabonnie: Good point, I think I like that better because we can have stricter checks if we want, e.g., we received more messages than expected.

test/run_e2e_test.py
dabonnie: Right, this wait is longer than ```TIMEOUT_SECONDS``` on line 210 in order to let spin complete. I'll comment so that's clear.

test/run_e2e_test.py
piraka9011: Python's `any()` should do you well:
```python
any(v == 0 for _, v in dict.iteritems())
```
Or something along those lines

Issue #103, opened 2020-03-04
Refactor LinuxProcessCpuMeasurementTestFixture Tests


dabonnie:
## Description

Given issues found in #96 and #83, refactor the ```LinuxProcessCpuMeasurementTestFixture``` tests to use the same utilities as the bugs' proposed fixes.

## Test Plan

* Unit test for new classes

## Documentation Plan

N/A

## Release Plan

* check in PR

## Acceptance Criteria

* [ ] Code has been implemented, reviewed and merged.
* [ ] Test plan has been completed
* [ ] Release plan has been completed

Once all above items are checked, this story can be moved to done.

## Resources

* #83 
* #96

PR #107, opened 2020-03-05
Refactor Linux process CPU measurement tests

dabonnie:
Moves #103 to done

### DIFF
--- a/system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
+++ b/system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
@@ -37,6 +37,9 @@
 #include "test_constants.hpp"
 #include "test_functions.hpp"
 
+
+namespace
+{
 using lifecycle_msgs::msg::State;
 using metrics_statistics_msgs::msg::MetricsMessage;
 using metrics_statistics_msgs::msg::StatisticDataPoint;
@@ -45,10 +48,8 @@ using moving_average_statistics::MovingAverageStatistics;
 using moving_average_statistics::StatisticData;
 using test_constants::kProcPidSamples;
 
-namespace
-{
 constexpr const char kTestProcessCpuNodeName[] = "test_measure_linux_process_cpu";
-}
+}  // namespace
 
 class MockLinuxProcessCpuMeasurementNode : public system_metrics_collector::
   LinuxProcessCpuMeasurementNode, public test_functions::PromiseSetter
@@ -153,7 +154,7 @@ TEST_F(LinuxProcessCpuMeasurementTestFixture, TestGetMetricName) {
 TEST_F(LinuxProcessCpuMeasurementTestFixture, TestManualMeasurement)
 {
   // first measurement caches
-  double cpu_active_percentage = test_measure_linux_process_cpu_->PeriodicMeasurement();
+  auto cpu_active_percentage = test_measure_linux_process_cpu_->PeriodicMeasurement();
   ASSERT_TRUE(std::isnan(cpu_active_percentage));
   // second measurement compares current and cached
   cpu_active_percentage = test_measure_linux_process_cpu_->PeriodicMeasurement();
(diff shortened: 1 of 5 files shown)


dabonnie:
Currently passing tests at iteration >1800

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -27,26 +27,31 @@
...
 #include "test_constants.hpp"
+#include "test_functions.hpp"
 
 using lifecycle_msgs::msg::State;
 using metrics_statistics_msgs::msg::MetricsMessage;
 using metrics_statistics_msgs::msg::StatisticDataPoint;
 using metrics_statistics_msgs::msg::StatisticDataType;
+using moving_average_statistics::MovingAverageStatistics;
 using moving_average_statistics::StatisticData;
+using test_constants::kProcPidSamples;
```
zmichaels11: It doesn't matter much since this is a `.cpp` file, but these can be moved into the anonymous namespace to explicitly restrict their scope to this file.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -60,6 +65,12 @@ class MockLinuxProcessCpuMeasurementNode : public system_metrics_collector::
    */
   double PeriodicMeasurement() override
   {
+    // first measurement does not add to the sample count
+    // proc cpu data requires two samples to compute a single measurement
+    if (++count_ > 1) {
```
zmichaels11: This will be true if count is >= 0. Is this intentional?

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -252,122 +129,187 @@ class LinuxProcessCpuMeasurementTestFixture : public ::testing::Test
...
+  ASSERT_EQ(
+    std::to_string(
+      pid) + "_cpu_percent_used", test_measure_linux_process_cpu_->GetMetricName());
 }
 
 TEST_F(LinuxProcessCpuMeasurementTestFixture, TestManualMeasurement)
 {
   // first measurement caches
-  double cpu_active_percentage = test_node_->PeriodicMeasurement();
+  double cpu_active_percentage = test_measure_linux_process_cpu_->PeriodicMeasurement();
```
zmichaels11: `const auto` unless you're performing implicit casting. In which case, please explicitly cast.

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
zmichaels11: sorry I mistread the code;
`++count > 1` does:
* preinc and return
* compare gt 1

so the first iteration its checking `1 > 1` which is `false`
and on the second iteration its checking `2 > 1` which is `true`

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
```diff
@@ -77,147 +88,12 @@ class MockLinuxProcessCpuMeasurementNode : public system_metrics_collector::
 private:
   system_metrics_collector::ProcPidCpuData MakeSingleMeasurement() override
   {
-    EXPECT_GT(test_constants::kProcPidSamples.size(), measurement_index_);
-    return test_constants::kProcPidSamples[measurement_index_++];
+    EXPECT_GT(kProcPidSamples.size(), measurement_index_);
```
zmichaels11:
```suggestion
    ASSERT_GT(kProcPidSamples.size(), measurement_index_);
```
Since the following line will index out of bounds if the `EXPECT_EQ` fails.

piraka9011: Should this be an `ASSERT` to make sure the next line never goes OOB?

zmichaels11: beat you to it by 7 minutes

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
dabonnie: Can use auto but not const (see line 160).

system_metrics_collector/test/system_metrics_collector/test_linux_process_cpu_measurement_node.cpp
dabonnie: Could use this if the return type was void.

Issue #110, opened 2020-03-12
Refactor Topic Statistics Measurements to use rcl time


dabonnie:
## Description

Instead of using rclcpp::Time, refactor topic statistics measurements to use [rcl_time_point_value_t](https://github.com/ros2/rcl/blob/master/rcl/include/rcl/time.h#L141) to represent [nanoseconds](http://docs.ros2.org/eloquent/api/rclcpp/classrclcpp_1_1Time.html#a0bab304d7b9e92dc568b069684275fe5). This removes the direct dependency on rclcpp and removes the responsibility of determining the clock type from the measurement classes.

This requires an update to the topic statistics interface, where upon a message received the time the message was received needs to be provided.

## Test Plan

* Unit test for new types

## Documentation Plan

* update class documentation to note new types and clock responsibility

## Release Plan

* check in PR

## Acceptance Criteria

* [ ] Code has been implemented, reviewed and merged.
* [ ] Test plan has been completed
* [ ] Release plan has been completed

Once all above items are checked, this story can be moved to done.

PR #113, opened 2020-03-12
Use rcutils_time_point_value_t instead of rclcpp::Time

dabonnie:
Moves #110 to done

### DIFF
--- a/system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
+++ b/system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
@@ -37,7 +37,7 @@ struct HasHeader<M, decltype((void) M::header)>: std::true_type {};
 template<typename M, typename Enable = void>
 struct TimeStamp
 {
-  static uint64_t value(const M & m)
+  static int64_t value(const M & m)
   {
     (void)m;
     return 0;
@@ -47,10 +47,10 @@ struct TimeStamp
 template<typename M>
 struct TimeStamp<M, typename std::enable_if<HasHeader<M>::value>::type>
 {
-  static uint64_t value(const M & m)
+  static int64_t value(const M & m)
   {
     auto stamp = m.header.stamp;
-    return stamp.nanosec;
+    return RCL_S_TO_NS(static_cast<int64_t>(stamp.sec)) + stamp.nanosec;
   }
 };
 }  // namespace
@@ -79,10 +79,11 @@ public:
   * Handle a new incoming message. Calculate message age if a valid Header is present.
   *
   * @param received_message, the message to calculate age of.
+  * @param time the message was received in nanoseconds
   */
   void OnMessageReceived(
     const T & received_message,
-    const rcl_time_point_value_t & now_nanoseconds) override
+    const rcl_time_point_value_t now_nanoseconds) override
   {
     const auto timestamp_from_header = TimeStamp<T>::value(received_message);
 
@@ -91,7 +92,7 @@ public:
       const auto age_millis = std::chrono::duration_cast<std::chrono::milliseconds>(age_nanos);
 
       system_metrics_collector::Collector::AcceptData(static_cast<double>(age_millis.count()));
-    }
+    }  // else no valid time to compute age
   }
 
 protected:
--- a/system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
+++ b/system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
@@ -18,13 +18,42 @@
 #include <chrono>
 #include <string>
 #include <sstream>
+#include <type_traits>
 
 #include "topic_statistics_collector.hpp"
 
-#include "message_filters/message_traits.h"  // TODO(dabonnie): remove and redefine here
 #include "rcl/time.h"
 #include "rcutils/logging_macros.h"
 
+
+namespace
+{
+template<typename M, typename = void>
+struct HasHeader : public std::false_type {};
+
+template<typename M>
+struct HasHeader<M, decltype((void) M::header)>: std::true_type {};
+
+template<typename M, typename Enable = void>
+struct TimeStamp
+{
+  static uint64_t value(const M & m)
+  {
+    (void)m;
+    return 0;
+  }
+};
+
+template<typename M>
+struct TimeStamp<M, typename std::enable_if<HasHeader<M>::value>::type>
+{
+  static uint64_t value(const M & m)
+  {
+    auto stamp = m.header.stamp;
+    return stamp.nanosec;
+  }
+};
+}  // namespace
 
 namespace topic_statistics_collector
 {
@@ -41,7 +70,7 @@ public:
    * Construct a ReceivedMessageAgeCollector object.
    *
    */
-  explicit ReceivedMessageAgeCollector()
+  ReceivedMessageAgeCollector()
   {}
 
   virtual ~ReceivedMessageAgeCollector() = default;
@@ -53,16 +82,13 @@ public:
   */
   void OnMessageReceived(const T & received_message, const uint64_t & now_nanoseconds) override
   {
-    const auto timestamp_from_header = message_filters::message_traits::TimeStamp<T>::value(
-      received_message);
+    const auto timestamp_from_header = TimeStamp<T>::value(received_message);
 
-    if (timestamp_from_header.nanoseconds()) {
+    if (timestamp_from_header && now_nanoseconds) {
+      const std::chrono::nanoseconds age_nanos{now_nanoseconds - timestamp_from_header};
+      const auto age_millis = std::chrono::duration_cast<std::chrono::milliseconds>(age_nanos);
 
-        const std::chrono::nanoseconds age_nanos{
-          now_nanoseconds - timestamp_from_header.nanoseconds()};
-        const auto age_millis = std::chrono::duration_cast<std::chrono::milliseconds>(age_nanos);
-
-        system_metrics_collector::Collector::AcceptData(static_cast<double>(age_millis.count()));
+      system_metrics_collector::Collector::AcceptData(static_cast<double>(age_millis.count()));
     }
   }
 
@@ -76,7 +102,6 @@ protected:
   {
     return true;
   }
-
 };
 
 }  // namespace topic_statistics_collector
--- a/system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
+++ b/system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
@@ -19,6 +19,7 @@
 #include <string>
 #include <sstream>
 #include <type_traits>
+#include <utility>
 
 #include "topic_statistics_collector.hpp"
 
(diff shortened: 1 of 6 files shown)


system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -56,41 +80,18 @@ class ReceivedMessageAgeCollector : public TopicStatisticsCollector<T>
   *
   * @param received_message, the message to calculate age of.
   */
-  void OnMessageReceived(const T & received_message) override
+  void OnMessageReceived(
+    const T & received_message,
+    const rcl_time_point_value_t & now_nanoseconds) override
```
zmichaels11: `rcl_time_point_value_t` is a `ulong64` so just pass by value.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -56,41 +80,18 @@ class ReceivedMessageAgeCollector : public TopicStatisticsCollector<T>
...
-        warn_msg <<
-          "Message header and current clock have different time sources, " <<
-          "cannot measure message age" <<
-          "\nmessage header clock type: " << timestamp_from_header.get_clock_type() <<
-          "\ncollector clock type: " << clock_.get_clock_type();
-        RCUTILS_LOG_WARN_NAMED("topic_statistics_collector", "%s", warn_msg.str().c_str());
-      }
-    }
-  }
+    const auto timestamp_from_header = TimeStamp<T>::value(received_message);
```
mm318: What is this template magic trying to do?

mm318: Is this what the circular dependency involving `message_filters` was needed for? And I am guessing that [this](https://github.com/ros-tooling/system_metrics_collector/pull/113/files#diff-5d4109d4db4bc0aeccb452eada7901beR29-R56) is copied and pasted from `message_filters` to break that circular dependency? Is there some way to additionally simplify this?

mm318: > What is this template magic trying to do?

Looks like the logic here is if message type `T` has a field called `header`, then the timestamp will be taken from that `header`, otherwise the timestamp will simply be 0. And later, if `timestamp_from_header == 0`, you won't accept the data.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -18,15 +18,43 @@
...
+  }
+};
+
+template<typename M>
+struct TimeStamp<M, typename std::enable_if<HasHeader<M>::value>::type>
+{
+  static uint64_t value(const M & m)
+  {
+    auto stamp = m.header.stamp;
+    return stamp.nanosec;
```
mm318: Does `stamp` not have a `sec` component that you're dropping?

dabonnie: It does, but if nanoseconds is set do we need it?

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
dabonnie: Yes, you are correct. If a header is present we attempt to extract a valid value. Notice how this doesn't return rclcpp::Time, but the rcl type (unlike [message_filters](https://github.com/ros2/message_filters/blob/master/include/message_filters/message_traits.h#L89)).

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
mm318: Yeah, you do. The `sec` field is not the same thing as the `nanosec` field just in a different unit. They have a `sec` field to hold the value for when the `nanosec` field rolls over.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
mm318: I think it's better to refactor `TimeStamp::value()` into standalone templated function(s), move those function into this class as private helper function(s) (looking like [this](https://stackoverflow.com/a/13402245)), move the `AcceptData()` logic into this private helper function(s). Finally, have `OnMessageReceived()` call this private helper function. (`OnMessageReceived()` can't become templated function(s) itself because it's virtual.)

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
```diff
@@ -63,34 +59,23 @@ class ReceivedMessagePeriodCollector : public TopicStatisticsCollector<T>
    * a lock to prevent race conditions when setting the time_last_message_received_ member.
    *
    * @param received_message
+   * @param time the message was received in nan
```
piraka9011: nanoseconds?

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
emersonknapp: e.g. nanosec <= 999,999,999

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
```diff
@@ -113,19 +98,14 @@ class ReceivedMessagePeriodCollector : public TopicStatisticsCollector<T>
...
-  rclcpp::Clock clock_;
-  /**
-   * Default uninitialized time. In order to compare rclcpp::Time they must come from
-   * the same type of clock.
+   * Default uninitialized time.
    */
-  rclcpp::Time uninitialized_time_;
-  rclcpp::Time time_last_message_received_ RCPPUTILS_TSA_GUARDED_BY(mutex_);
+  rcl_time_point_value_t time_last_message_received_{kUninitializedTime}
+  RCPPUTILS_TSA_GUARDED_BY(mutex_);
```
piraka9011: `mutex_` is defined on the next line. Is that fine?

dabonnie: I'm not aware of any issues with that

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
dabonnie: I'm not seeing the advantage, especially since it seems strange to branch the header extraction compared to doing it by example.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -18,15 +18,43 @@
...
+template<typename M, typename = void>
+struct HasHeader : public std::false_type {};
+
+template<typename M>
+struct HasHeader<M, decltype((void) M::header)>: std::true_type {};
+
+template<typename M, typename Enable = void>
+struct TimeStamp
+{
+  static int64_t value(const M & m)
```
thomas-moulard:
```suggestion
  static int64_t value(const M &)
```

and remove `(void)m`.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -18,15 +18,43 @@
...
 
-#include "message_filters/message_traits.h"
 #include "rcl/time.h"
-#include "rclcpp/clock.hpp"
 #include "rcutils/logging_macros.h"
 
 
+namespace
+{
+template<typename M, typename = void>
```
thomas-moulard: Please add documentation for this complex piece of C++.

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
```diff
@@ -23,9 +23,13 @@
 #include "system_metrics_collector/collector.hpp"
 
 #include "rcl/time.h"
-#include "rclcpp/clock.hpp"
 
 
+namespace
+{
+constexpr const int64_t kUninitializedTime{0};
```
thomas-moulard: need to be defined in `received_message_period.cpp`

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -18,15 +18,43 @@
...
+struct HasHeader : public std::false_type {};
+
+template<typename M>
+struct HasHeader<M, decltype((void) M::header)>: std::true_type {};
+
+template<typename M, typename Enable = void>
```
thomas-moulard: ditto doc here

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
thomas-moulard:
```suggestion
static constexpr const int64_t kUninitializedTime{0};
```

zmichaels11: I think `static` is redundant here since the anonymous namespace is going to make it file-scope.

zmichaels11: IMO: this should instead be moved into `namespace topic_statistics_collector`

Because `kUninitializedTime` is a numeric type, duplication in the code shouldn't matter, especially since this is a private header. I'm not fully confident with this second opinion so I'm going to double check it.

zmichaels11: I looked up the first comment, we should not use [anonymous namepaces](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-unnamed) in headers.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
dabonnie: That's neat.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -18,15 +18,43 @@
...
 
 #include "topic_statistics_collector.hpp"
 
-#include "message_filters/message_traits.h"
 #include "rcl/time.h"
-#include "rclcpp/clock.hpp"
 #include "rcutils/logging_macros.h"
 
 
+namespace
```
zmichaels11: Anonymous namespaces should not be used in a header.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
mm318: My suggestion also brings up another question. What if message `T` has a field called `header`, but the timestamp is actually 0? Who is setting these headers? Can everybody be trusted to not start (simulation) time at 0?

thomas-moulard: Good point, if you feel it's worth it, we could return `std::pair(int64_t, bool /* is_initialized */)` which is an OK alternative to `std::optional` we can't use yet.

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
thomas-moulard: yes, that too.

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
```diff
@@ -40,19 +44,11 @@ class ReceivedMessagePeriodCollector : public TopicStatisticsCollector<T>
...
-   * the same clock type.
+   * Construct a ReceivedMessagePeriodCollector object.
    *
-   * @param clock input clock to use in order to measure received message
-   * period, default is RCL_STEADY_TIME
    */
-  explicit ReceivedMessagePeriodCollector(
-    const rclcpp::Clock & clock = rclcpp::Clock{RCL_STEADY_TIME})
-  : clock_{clock}
+  ReceivedMessagePeriodCollector()
```
thomas-moulard: I'd call the base class ctor explicitly, for readability.

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
dabonnie: That was my thought as well concerning static.

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
```diff
@@ -23,12 +23,13 @@
 #include "system_metrics_collector/collector.hpp"
 
 #include "rcl/time.h"
-#include "rclcpp/clock.hpp"
 
 
 namespace topic_statistics_collector
 {
 
+static constexpr const int64_t kUninitializedTime{0};
```
zmichaels11: I think `static` is redundant in this context

thomas-moulard: Variables should never be marked as static in headers, they would be duplicated in each compilation header.

dabonnie: Resolved

PR #115, opened 2020-03-15
Add statistic name and unit interfaces

prajakta-gokhale:
Changes to `topic_statistics_collector` interface. Needed for https://github.com/ros-tooling/system_metrics_collector/pull/112.

### DIFF
--- a/system_metrics_collector/src/system_metrics_collector/collector.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.hpp
@@ -18,6 +18,7 @@
 #include <mutex>
 #include <string>
 
+#include "metric_details_interface.hpp"
 #include "moving_average_statistics/moving_average.hpp"
 #include "moving_average_statistics/types.hpp"
 
@@ -29,7 +30,7 @@ namespace system_metrics_collector
 /**
  * Simple class in order to collect observed data and generate statistics for the given observations.
  */
-class Collector
+class Collector : virtual public MetricDetailsInterface
 {
 public:
   Collector() = default;
@@ -90,6 +91,20 @@ public:
    * @return true if stopped, false if an error occurred
    */
   virtual bool Stop();
+
+  /**
+   * Return the name of the metric collected.
+   *
+   * @return string reporesenting name of the collected metric
+   */
+  std::string GetMetricName() const override = 0;
+
+  /**
+   * Return the measurement unit of the metric collected.
+   *
+   * @return string representing unit of the metric collected
+   */
+  std::string GetMetricUnit() const override = 0;
 
 private:
   /**
--- a/system_metrics_collector/src/system_metrics_collector/collector.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.hpp
@@ -30,7 +30,7 @@ namespace system_metrics_collector
 /**
  * Simple class in order to collect observed data and generate statistics for the given observations.
  */
-class Collector : virtual public MetricDetailsInterface
+class Collector : public MetricDetailsInterface
 {
 public:
   Collector() = default;
--- a/system_metrics_collector/src/system_metrics_collector/collector.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/collector.hpp
@@ -92,20 +92,6 @@ public:
    */
   virtual bool Stop();
 
-  /**
-   * Return the name of the metric collected.
-   *
-   * @return string reporesenting name of the collected metric
-   */
-  std::string GetMetricName() const override = 0;
-
-  /**
-   * Return the measurement unit of the metric collected.
-   *
-   * @return string representing unit of the metric collected
-   */
-  std::string GetMetricUnit() const override = 0;
-
 private:
   /**
    * Override in order to perform necessary starting steps.
--- a/system_metrics_collector/src/system_metrics_collector/metric_details_interface.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/metric_details_interface.hpp
@@ -1,0 +1,38 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef SYSTEM_METRICS_COLLECTOR__METRIC_DETAILS_INTERFACE_HPP_
+#define SYSTEM_METRICS_COLLECTOR__METRIC_DETAILS_INTERFACE_HPP_
+
+#include <string>
+
+namespace system_metrics_collector
+{
+/**
+ * Class to represent a single metric for collection or publication.
+ */
+
+class MetricDetailsInterface
+{
+public:
+  virtual ~MetricDetailsInterface() = default;
+
+  virtual std::string GetMetricName() const = 0;
+
+  virtual std::string GetMetricUnit() const = 0;
+};
+
+}  // namespace system_metrics_collector
+
+#endif  // SYSTEM_METRICS_COLLECTOR__METRIC_DETAILS_INTERFACE_HPP_
--- a/system_metrics_collector/src/system_metrics_collector/metric_details_interface.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/metric_details_interface.hpp
@@ -20,7 +20,8 @@
 namespace system_metrics_collector
 {
 /**
- * Class to represent a single metric for collection or publication.
+ * Interface to represent a single metric's name and unit,
+ * which are used for metric message generation and publication.
  */
 
 class MetricDetailsInterface
@@ -28,8 +29,18 @@ class MetricDetailsInterface
 public:
   virtual ~MetricDetailsInterface() = default;
 
+  /**
+   * Return a single metric's name.
+   *
+   * @return a string representing the metric name
+   */
   virtual std::string GetMetricName() const = 0;
 
+  /**
+   * Return a single metric's measurement unit.
+   *
+   * @return a string representing the metric unit
+   */
   virtual std::string GetMetricUnit() const = 0;
 };
 
--- a/system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
@@ -21,6 +21,7 @@
 #include "builtin_interfaces/msg/time.hpp"
 #include "metrics_statistics_msgs/msg/metrics_message.hpp"
 
+#include "metric_details_interface.hpp"
 #include "moving_average_statistics/types.hpp"
 
 namespace system_metrics_collector
@@ -49,7 +50,7 @@ metrics_statistics_msgs::msg::MetricsMessage GenerateStatisticMessage(
 /**
  * Simple class to facilitate publishing messages containing statistics data
  */
-class MetricsMessagePublisher
+class MetricsMessagePublisher : virtual public MetricDetailsInterface
 {
 public:
   /**
@@ -63,14 +64,14 @@ public:
    *
    * @return a string of the name for this measured metric
    */
-  virtual std::string GetMetricName() const = 0;
+  std::string GetMetricName() const override = 0;
 
   /**
    * Returns the name of the measurement unit of this metric
    *
    * @return a string of the name of the measurement unit of this metric
    */
-  virtual const std::string & GetMetricUnit() const = 0;
+  std::string GetMetricUnit() const override = 0;
 };
 }  // namespace system_metrics_collector
 
--- a/system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
+++ b/system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
@@ -21,7 +21,6 @@
 #include "builtin_interfaces/msg/time.hpp"
 #include "metrics_statistics_msgs/msg/metrics_message.hpp"
 
-#include "metric_details_interface.hpp"
 #include "moving_average_statistics/types.hpp"
 
 namespace system_metrics_collector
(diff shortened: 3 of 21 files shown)


system_metrics_collector/src/topic_statistics_collector/topic_statistics_collector.hpp
```diff
@@ -35,12 +35,26 @@ class TopicStatisticsCollector : public system_metrics_collector::Collector
...
    * @tparam T the ROS2 message type to collect
    */
   virtual void OnMessageReceived(const T & received_message) = 0;
+
+  /**
+   * Return the name to use for collected statistic
+   *
+   * @return a string of the name for this statistic
+   */
+  virtual const std::string GetStatisticName() const = 0;
```
dabonnie: These duplicate the interface found in ::MetricsMessagePublisher. Instead of duplicating code, we should consider using the interface. If the publish method is too restrictive (I think it is), now is the time to remove it from the interface definition.

system_metrics_collector/src/topic_statistics_collector/constants.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <string>
+
+namespace topic_statistics_collector
+{
+
+namespace topic_statistics_constants
+{
+constexpr const char kMsgAgeStatName[] = "message_age";
+constexpr const char kMsgPeriodStatName[] = "message_period";
+constexpr const char kMillisecondUnitName[] = "ms";
```
zmichaels11: constant strings need to be defined in the matching cpp file.

system_metrics_collector/src/topic_statistics_collector/topic_statistics_collector.hpp
prajakta-gokhale: Yeah it is very similar to the interfaces in `MetricsMessagePublisher`. Instead of removing the `PublishStatisticMessage()` method from there, would it make sense to separate the metric name and unit methods into a different interface? Then the `Collector` and `MetricsMessagePublisher` can both inherit it.

dabonnie: That sounds good.

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -82,6 +83,25 @@ class ReceivedMessageAgeCollector : public TopicStatisticsCollector<T>
     }
   }
 
+  /**
+   * Return message age metric name
+   *
+   * @return a string representing message age metric name
+   */
+  const std::string GetStatisticName() const override
```
thomas-moulard:
```suggestion
 std::string GetStatisticName() const override
```

system_metrics_collector/src/topic_statistics_collector/received_message_age.hpp
```diff
@@ -82,6 +83,25 @@ class ReceivedMessageAgeCollector : public TopicStatisticsCollector<T>
...
+  {
+    return topic_statistics_constants::kMsgAgeStatName;
+  }
+  /**
+   * Return messge age metric unit
+   *
+   * @return a string representing messager age metric unit
+   */
+  const std::string GetStatisticUnit() const override
```
thomas-moulard:
```suggestion
 std::string GetStatisticUnit() const override
```

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
```diff
@@ -80,6 +81,25 @@ class ReceivedMessagePeriodCollector : public TopicStatisticsCollector<T>
     }
   }
 
+  /**
+   * Return message period metric name
+   *
+   * @return a string representing message period metric name
+   */
+  const std::string GetStatisticName() const override
```
thomas-moulard:
```suggestion
 std::string GetStatisticName() const override
```

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
```diff
@@ -80,6 +81,25 @@ class ReceivedMessagePeriodCollector : public TopicStatisticsCollector<T>
...
+  {
+    return topic_statistics_constants::kMsgPeriodStatName;
+  }
+  /**
+   * Return message period metric unit
+   *
+   * @return a string representing message period metric name
+   */
+  const std::string GetStatisticUnit() const override
```
thomas-moulard:
```suggestion
  std::string GetStatisticUnit() const override
```

system_metrics_collector/src/topic_statistics_collector/topic_statistics_collector.hpp
thomas-moulard:
```suggestion
  virtual std::string GetStatisticName() const = 0;
```

system_metrics_collector/src/topic_statistics_collector/received_message_period.hpp
```diff
@@ -80,6 +81,25 @@ class ReceivedMessagePeriodCollector : public TopicStatisticsCollector<T>
...
+   * @return a string representing message period metric name
+   */
+  const std::string GetStatisticName() const override
+  {
+    return topic_statistics_constants::kMsgPeriodStatName;
+  }
+  /**
+   * Return message period metric unit
+   *
+   * @return a string representing message period metric name
```
thomas-moulard: unit or name? the `@return` seems wrong

system_metrics_collector/test/topic_statistics_collector/test_received_message_age.cpp
```diff
@@ -177,3 +178,14 @@ TEST(ReceivedMessageAgeTest, TestAgeMeasurement) {
...
+TEST(ReceivedMessageAgeTest, TestGetStatNameAndUnit) {
+  TestReceivedMessageAgeCollector test_collector{};
+
+  EXPECT_EQ(
+    test_collector.GetStatisticName(),
+    topic_statistics_collector::topic_statistics_constants::kMsgAgeStatName);
+  EXPECT_EQ(
+    test_collector.GetStatisticUnit(),
+    topic_statistics_collector::topic_statistics_constants::kMillisecondUnitName);
+}
```
thomas-moulard: This is the definition of a change detector test https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

Let's test the string is not empty instead, or any other useful, universally true about those metrics names and units.

system_metrics_collector/test/topic_statistics_collector/test_received_message_period.cpp
```diff
@@ -121,3 +122,14 @@ TEST(ReceivedMessagePeriodTest, TestPeriodMeasurement) {
...
+TEST(ReceivedMessagePeriodTest, TestGetStatNameAndUnit) {
+  TestReceivedMessagePeriodCollector test_collector{};
+
+  EXPECT_EQ(
+    test_collector.GetStatisticName(),
+    topic_statistics_collector::topic_statistics_constants::kMsgPeriodStatName);
+  EXPECT_EQ(
+    test_collector.GetStatisticUnit(),
+    topic_statistics_collector::topic_statistics_constants::kMillisecondUnitName);
+}
```
thomas-moulard: ditto

system_metrics_collector/test/topic_statistics_collector/test_received_message_age.cpp
prajakta-gokhale: Good to know! Thanks for the link :)

system_metrics_collector/src/topic_statistics_collector/constants.hpp
prajakta-gokhale: This pattern also exists in this package [elsewhere](https://github.com/ros-tooling/system_metrics_collector/blob/master/system_metrics_collector/src/system_metrics_collector/constants.hpp). I'll address this comment everywhere in a separate PR to keep the diff short here.

system_metrics_collector/src/system_metrics_collector/metric_details_interface.hpp
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Class to represent a single metric for collection or publication.
+ */
+
+class MetricDetailsInterface
+{
+public:
+  virtual ~MetricDetailsInterface() = default;
+
+  virtual std::string GetMetricName() const = 0;
```
dabonnie: Doc for usage?

system_metrics_collector/src/system_metrics_collector/metric_details_interface.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+  virtual std::string GetMetricUnit() const = 0;
```
dabonnie: Doc for usage?

system_metrics_collector/src/system_metrics_collector/metric_details_interface.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#ifndef SYSTEM_METRICS_COLLECTOR__METRIC_DETAILS_INTERFACE_HPP_
+#define SYSTEM_METRICS_COLLECTOR__METRIC_DETAILS_INTERFACE_HPP_
+
+#include <string>
+
+namespace system_metrics_collector
+{
+/**
+ * Class to represent a single metric for collection or publication.
```
dabonnie:
```suggestion
 * Interface to represent a single metric's name and units, which are used for message generation and publication.
```

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
```diff
@@ -49,7 +50,7 @@ metrics_statistics_msgs::msg::MetricsMessage GenerateStatisticMessage(
 /**
  * Simple class to facilitate publishing messages containing statistics data
  */
-class MetricsMessagePublisher
+class MetricsMessagePublisher : virtual public MetricDetailsInterface
```
dabonnie: Why is ```virtual``` needed? Here, you shouldn't need to extend ```MetricDetailsInterface``` because Collector->PeriodicMeasurementNode, Collector->TopicStatisticsCollector exposes the metric details.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
```diff
@@ -63,14 +64,14 @@ class MetricsMessagePublisher
    *
    * @return a string of the name for this measured metric
    */
-  virtual std::string GetMetricName() const = 0;
+  std::string GetMetricName() const override = 0;
```
dabonnie: These shouldn't be needed here anymore given the Collector changes.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
```diff
@@ -93,6 +93,20 @@ class PeriodicMeasurementNode : public system_metrics_collector::Collector,
   rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn on_error(
     const rclcpp_lifecycle::State & previous_state);
 
+  /**
+   * Return the name of the metric collected.
+   *
+   * @return string reporesenting name of the collected metric
+   */
+  std::string GetMetricName() const override = 0;
```
dabonnie: This is really confusing. PeriodicMeasurementNode inherits from both Collector and MetricsMessagePublisher.

system_metrics_collector/src/system_metrics_collector/collector.hpp
```diff
@@ -29,7 +30,7 @@ namespace system_metrics_collector
 /**
  * Simple class in order to collect observed data and generate statistics for the given observations.
  */
-class Collector
+class Collector : virtual public MetricDetailsInterface
```
dabonnie: I think this is a good change, but leaves us with some cleanup. See other comments.

prajakta-gokhale: See comment about virtual inheritance.

system_metrics_collector/src/system_metrics_collector/periodic_measurement_node.hpp
prajakta-gokhale: That is why [Virtual Inheritance](https://en.wikipedia.org/wiki/Virtual_inheritance) is used for disambiguation.

system_metrics_collector/src/system_metrics_collector/metrics_message_publisher.hpp
prajakta-gokhale: I didn't remove it from here because I'm not sure there won't be a use-case where a class only inherits from `MetricsMessagePublisher` and not `Collector`. In the spirit of [YAGNI](https://martinfowler.com/bliki/Yagni.html) principle, I'm okay removing it.

system_metrics_collector/src/topic_statistics_collector/constants.hpp
thomas-moulard: +1 - let's make sure this gets fixed, as this could cause "invisible issues" as we don't have a CI attempting to compile in release, for instance.

thomas-moulard:
The coverage seems to be decreasing with this PR, is there any easy fix for us to try not to reduce coverage on this?

prajakta-gokhale:
> The coverage seems to be decreasing with this PR, is there any easy fix for us to try not to reduce coverage on this?

After rebasing on `origin/master` it seems to be decreasing by `0.17%`. The only new code added is an interface whose implementation is unit tested in concrete derived classes.

Increasing it would be a non-trivial change.

thomas-moulard:
No worries, this is fine. On the report:

https://codecov.io/gh/ros-tooling/system_metrics_collector/pull/115/diff?src=pr&el=tree-more

`test_collector.cpp` `GetMetricName/Unit` methods are not hit currently, just FYI.

system_metrics_collector/test/system_metrics_collector/test_periodic_measurement_node.cpp
```diff
@@ -95,10 +95,9 @@ class TestPeriodicMeasurementNode : public ::system_metrics_collector::PeriodicM
     return kTestMetricName;
   }
 
-  const std::string & GetMetricUnit() const override
+  std::string GetMetricUnit() const override
   {
-    static const std::string unit_name{kTestMetricUnit};
-    return unit_name;
+    return kTestMetricUnit;
```
dabonnie: This is a good example of the misses increasing. Since test files are included, lines 98 and 100 are missed because there's nothing calling them that is tested.

system_metrics_collector/src/topic_statistics_collector/constants.hpp
dabonnie: I'll file an issue.

dabonnie: Filed https://github.com/ros-tooling/system_metrics_collector/issues/116

Issue #120, opened 2020-03-31
Use a dummy message instead of sensor_msgs in tests and demos


prajakta-gokhale:
can we avoid this by using a dummy message in the test folder?

_Originally posted by @thomas-moulard in https://github.com/ros-tooling/system_metrics_collector/pull/112_

PR #121, opened 2020-04-01
Replace IMU messages with new dummy messages

prajakta-gokhale:
* Define a new message type `DummyMessage` that has a `Header` 
* Use `DummyMessage` in unit tests instead of `sensor_msgs::msg::Imu` 
* Remove dependency on `sensor_msgs`

Fixes https://github.com/ros-tooling/system_metrics_collector/issues/120.


prajakta-gokhale:
> Why not just put the message in test/dummy.msg, and drop the new package altogether?

I'll try that. Since it's needed in 2 packages (`libstatictics_collector` and `topic_statisticvs_collector`), is it okay to depend on `test/dummy.msg` of one package from another?

thomas-moulard:
I think duplicating the test message is OK as this is trivial.

prajakta-gokhale:
As I understand, in order to move `DummyMessage` to individual packages that need them,  these library changes had to be made:
* Create two separate target libraries `${PROJECT_NAME}` and `${PROJECT_NAME}_shared`. The message generation step creates a Utility library with the name `${PROJECT_NAME}`. This has to be kept separate from the Shared library that the package exports.
* The Utility library name has to be same as `${PROJECT_NAME}` because all the typesupport_impl's are named `package_name__typesupport__rmw_name` by rosidl generators.

libstatistics_collector/package.xml
```diff
@@ -12,11 +12,17 @@
...
+  <exec_depend>rosidl_default_runtime</exec_depend>
+  <exec_depend>std_msgs</exec_depend>
+
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
-  <test_depend>sensor_msgs</test_depend>
-  <test_depend>std_msgs</test_depend>
+
+  <member_of_group>rosidl_interface_packages</member_of_group>
```
thomas-moulard, line 25: needed?

libstatistics_collector/CMakeLists.txt
```diff
@@ -30,16 +30,23 @@ endif()
...
+
+rosidl_generate_interfaces(${PROJECT_NAME}
+  "test/msg/DummyMessage.msg"
+  DEPENDENCIES std_msgs
+)
 
 include_directories(include)
 
-add_library(${PROJECT_NAME} SHARED
+add_library(${PROJECT_NAME}_shared SHARED
```
thomas-moulard, line 43: I think you can remove `_shared` by passing `LIBRARY_NAME`to `rosdl_generate_interfaces`

https://github.com/ros2/rosidl/blob/master/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L39

libstatistics_collector/package.xml
prajakta-gokhale: Yes. Fails to compile otherwise (see [rosidl_generate_interfaces.cmake](https://github.com/ros2/rosidl/blob/66f9a0e92962778edddc2ce6522ec259d7a76331/rosidl_cmake/cmake/rosidl_generate_interfaces.cmake#L229-L232)).

libstatistics_collector/CMakeLists.txt
```diff
@@ -82,7 +92,7 @@ install(
   DESTINATION include
 )
 
-ament_export_dependencies(rcl rcpputils)
+ament_export_dependencies(rcl rcpputils rosidl_default_runtime)
```
thomas-moulard, line 95: alphasort?

prajakta-gokhale: It is alphasorted :)

Issue #130, opened 2020-04-08
Failure building on OSX


dabonnie:
## Description

Unable to build `libstatistic_collector` on OSX.

## Expected Behavior
OSX build passes

## Actual Behavior
Found when running CI via https://github.com/ros2/rclcpp/pull/1050

```
--- stderr: libstatistics_collector
10:51:27 /Users/osrf/jenkins-agent/workspace/ci_osx/ws/src/system_metrics_collector/libstatistics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp:30:51: error: constexpr variable cannot have non-literal type 'const std::array<double, kExpectedSize>'
10:51:27 constexpr const std::array<double, kExpectedSize> kTestData{-3.5, -2.1, -1.1, 0.0, 4.7, 5.0,
10:51:27                                                   ^
10:51:27 /Users/osrf/jenkins-agent/workspace/ci_osx/ws/src/system_metrics_collector/libstatistics_collector/test/moving_average_statistics/test_moving_average_statistics.cpp:30:51: error: implicit instantiation of undefined template 'std::__1::array<double, 9>'
10:51:27 /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/__tuple:223:64: note: template is declared here
10:51:27 template <class _Tp, size_t _Size> struct _LIBCPP_TEMPLATE_VIS array;
10:51:27                                                                ^
10:51:27 2 errors generated.
10:51:27 make[2]: *** [CMakeFiles/test_moving_average_statistics.dir/test/moving_average_statistics/test_moving_average_statistics.cpp.o] Error 1
10:51:27 make[1]: *** [CMakeFiles/test_moving_average_statistics.dir/all] Error 2
10:51:27 make[1]: *** Waiting for unfinished jobs....
10:51:27 make: *** [all] Error 2
10:51:27 ---
10:51:27 Failed   <<< libstatistics_collector	[ Exited with code 2 ]
```

https://ci.ros2.org/job/ci_osx/8133/console#console-section-5

## To Reproduce
PROMPT: Steps to reproduce the behavior:
1. Start a CI job for the `libstatistic_collector` on OSX

## System (please complete the following information)
 - OS: CI OSX
 - ROS2 Distro: master

emersonknapp:
Reviewed at Sprint #17 review

PR #132, opened 2020-04-08
Add OSX and Windows to build libstatistics_collector

dabonnie:
This moves #130 to done.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
         os: [macOS-latest, ubuntu-18.04, windows-latest]
     steps:
       - uses: ros-tooling/setup-ros@0.0.17
-      - uses: ros-tooling/action-ros-ci@0.0.13
+      - uses: ros-tooling/action-ros-ci@0.0.15
         with:
           package-name: libstatistics_collector
           colcon-mixin-name: coverage-gcc
@@ -35,7 +35,7 @@ jobs:
           os: [ubuntu-18.04]
     steps:
     - uses: ros-tooling/setup-ros@0.0.17
-    - uses: ros-tooling/action-ros-ci@0.0.13
+    - uses: ros-tooling/action-ros-ci@0.0.15
       with:
         package-name: libstatistics_collector metrics_statistics_msgs system_metrics_collector
         colcon-mixin-name: coverage-gcc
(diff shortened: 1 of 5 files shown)


dabonnie:
Codecov currently broken (known), Windows is failing due to [ASIO](https://index.ros.org/doc/ros2/Troubleshooting/#cmake-packages-unable-to-find-asio-tinyxml2-tinyxml-eigen-or-log4cxx). Unclear how to fix and we need Windows CI for the libstatistics_collector` package.

.github/workflows/test.yml
```diff
@@ -18,7 +35,7 @@ jobs:
           os: [ubuntu-18.04]
     steps:
     - uses: ros-tooling/setup-ros@0.0.17
-    - uses: ros-tooling/action-ros-ci@0.0.13
+    - uses: ros-tooling/action-ros-ci@0.0.15
       with:
         package-name: libstatistics_collector metrics_statistics_msgs system_metrics_collector
```
prajakta-gokhale, line 40: Do we need to remove `libstatistics_collector` from here now?

.github/workflows/test.yml
```diff
@@ -10,6 +10,23 @@ on:
     - cron:  '0 * * * *'
 
 jobs:
+  build_and_test_libstatistics_collector:
```
prajakta-gokhale, line 13: What about `metrics_statistics_msgs`? Since that will also be a ros2 core dependency, do we need to add the same testing for it also?

.github/workflows/test.yml
dabonnie: I left it here for the code coverage report.

.github/workflows/test.yml
dabonnie: Good point, I'll add it here (or we have a separate job)

prajakta-gokhale: Separate job would be better imo.

dabonnie:
> Let's document a plan to fix Windows failures, looks good otherwise.

I don't think there's a viable plan here unless we change `setup-ros`. I'll at lest file an issue there and I'm thinking we remove the Windows CI given its current flakiness.

piraka9011:
@dabonnie  This needs setup-ros@0.0.19

PR #138, opened 2020-04-18
Remove libstatistics_collector package

prajakta-gokhale:
Remove `libstatistics_collector` now that it is in a separate repo.


emersonknapp:
Does this need any extra infra to build against libstatistics_collector via package.xml or .repos or anything now that it isn't in the same repo?

prajakta-gokhale:
> Does this need any extra infra to build against libstatistics_collector via package.xml or .repos or anything now that it isn't in the same repo?

No. Since `libstatistics_collector` was already a separate package in this repo, it was already listed as a dependency of this package. Also with https://github.com/ros2/ros2/pull/897 merged, `libstatistics_collector` is part of ros2.repos so no need to supply it externally to the ACtions either.

